PR #78, opened 2020-03-22
✨ feat(typescript): add types of webhook definitions to use on other projects

dominguezcelada:
## 📝 Summary

Add `index.d.ts` to expose types for a parsed version of `index.json` from external projects

## ⛱ Motivation and Context

On the [migration to TypeScript](https://github.com/dominguezcelada/webhooks.js/pull/3) for [webhooks.js](https://github.com/dominguezcelada/webhooks.js) this type would be necessary to be used in files like [`typescript-validate.ts`](https://github.com/octokit/webhooks.js/blob/master/test/typescript-validate.ts)


gr2m:
Ah nevermind my comment, I mixed up packages 🤦‍♂

PR #98, opened 2020-05-14
feat: event descriptions

### LINKED ISSUE #89: extract event descriptions
Let's update the [section-to-webhook](https://github.com/octokit/webhooks/blob/40ceb12f15d74be4d0274028b5ad28aceb1ede77/lib/section-to-webhook.js#L17) to also extract webhook descriptions.

Example: https://developer.github.com/v3/activity/events/types/#checksuiteevent

Screen Shot 2020-05-04 at 1 27 25 PM

zeke:
This PR updates the build process to extract descriptions from the events listed at https://developer.github.com/v3/activity/events/types/

Notes:

- Supports multi-paragraph descriptions
- HTML is converted to Markdown using `turndown`, following prior art from other Octokit projects
- Links to `/v3` are rewritten as fully-qualified `developer.github.com` URLs
- Introduced a basic JSON schema and added `ajv` to validate the shape of each extracted webhook object.
- I refactored `section-to-webhook.js` to make it easier to read and refine. I was also able to remove some workarounds around quoted action names that were no longer necessary.
- Removed the workaround for `pull_request.synchronize` which is no longer needed.

cc @rachmari @sarahs @bswinnerton @jasonetco

Resolves #89

### DIFF
--- a/index.json
+++ b/index.json
@@ -5,6 +5,7 @@
     "actions": [
       "completed",
       "created",
+      "requested_action",
       "rerequested"
     ],
     "examples": [
@@ -13416,7 +13417,6 @@
       "edited",
       "labeled",
       "locked",
-      "merged",
       "opened",
       "ready_for_review",
       "reopened",
--- a/index.json
+++ b/index.json
@@ -1,6 +1,7 @@
 [
   {
     "name": "check_run",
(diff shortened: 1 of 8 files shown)


index.json
```diff
@@ -1,10 +1,10 @@
...
     "actions": [
       "completed",
       "created",
-      "requested_action",
```
gr2m, line 7: I'm looking into why this is happening

index.json
```diff
@@ -13380,12 +13409,14 @@
...
   {
     "name": "pull_request",
+    "description": "Triggered when a [pull request](https://developer.github.com/v3/pulls) is `assigned`, `unassigned`, `labeled`, `unlabeled`, `opened`, `edited`, `closed`, `reopened`, `synchronize`, `ready_for_review`, `locked`, `unlocked` or when a pull request review is requested or removed.",
     "actions": [
       "assigned",
       "closed",
       "edited",
       "labeled",
       "locked",
+      "merged",
```
zeke, line 13419: @gr2m this one got added to FYI

gr2m: ah thanks for the pointer, `merged` is not a valid action

rachmari:
@zeke I mentioned this to @gr2m and it sounded like it wouldn't impact this work significantly, but I've moved and reworked that page a bit in https://github.com/github/internal-developer.github.com/pull/6352. I'm hoping to ship that this week, but I can hold on shipping it if it will cause significant changes to your work here.

zeke:
Thanks for the heads-up @rachmari. I will defer to Gregor's judgement on that.

zeke:
That was quick! Thanks for the cleanup @gr2m 🙌

PR #137, opened 2020-08-03
fix(payload–examples): add "installation" field to pull_request-* payload examples

dominguezcelada:
## 📝 Summary

Add `"installation"` field to `pull_request.*` payload examples

## ⛱ Motivation and Context

Fixes https://github.com/octokit/webhooks.js/issues/176
Fix https://github.com/octokit/webhooks/issues/138


gr2m:
btw the emoji are fun in comments and descriptions, but I have to remove them from the beginning of commit messages because of semantic-release

PR #151, opened 2020-08-23
refactor(typescript): Migrate repo to TypeScript

### LINKED ISSUE #152: Migrate @octokit/webhooks-definitions to Typescript

dominguezcelada:
## 📝 Summary

WIP

## ⛱ Motivation and Context

Fixes #152 

## 📚 References:

Using Aribnb's utility https://github.com/airbnb/ts-migrate


gr2m:
Hmm I don't think it's worth migrating repos like this to TypeScript. It's just a hacky build repository that creates static assets, which then can be consumed by other libraries. There are methods being exported. I thinkTypeScript will get more in the way than it will help

dominguezcelada:
> Hmm I don't think it's worth migrating repos like this to TypeScript. It's just a hacky build repository that creates static assets, which then can be consumed by other libraries. There are methods being exported. I thinkTypeScript will get more in the way than it will help

I see. Closing this PR then. 

Still, the issue to migrate this repo (manually) to TypeScript is interesting right?

gr2m:
Let's leave it written in JavaScript, please

PR #177, opened 2020-09-29
fix: add installation object to pull_request_review and pull_request_review_comment events

roopakv:
The pull request review payloa has an installation object as defined here: https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request_review

Adding this will end up fixing the typescript definitions

### DIFF
--- a/index.json
+++ b/index.json
@@ -18535,6 +18535,10 @@
           "received_events_url": "https://api.github.com/users/Codertocat/received_events",
           "type": "User",
           "site_admin": false
+        },
+        "installation": {
+          "id": 1,
+          "node_id": "MDIzOkludGVncmF0aW9uSW5zdGFsbGF0aW9uMQ=="
         }
       },
       {
--- a/index.json
+++ b/index.json
@@ -18535,10 +18535,6 @@
           "received_events_url": "https://api.github.com/users/Codertocat/received_events",
           "type": "User",
           "site_admin": false
-        },
-        "installation": {
-          "id": 1,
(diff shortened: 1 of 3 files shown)


index.json
```diff
@@ -18535,6 +18535,10 @@
           "received_events_url": "https://api.github.com/users/Codertocat/received_events",
           "type": "User",
           "site_admin": false
+        },
+        "installation": {
+          "id": 1,
+          "node_id": "MDIzOkludGVncmF0aW9uSW5zdGFsbGF0aW9uMQ=="
```
gr2m, line 18541: the `index.json` file is actually generated, what you have to do is to add the `installation` to the payload examples, in this case to https://github.com/octokit/webhooks/blob/master/payload-examples/api.github.com/pull_request_review.submitted.payload.json. This should have been fixed by https://github.com/octokit/webhooks/pull/137 but looks like we missed a few. You can also add new payload examples in case the payloads are different for different actions

After updating the file, run `node bin/octokit-webhooks.js check --cached` which will update `index.json`.

gr2m (changes requested):

index.json
roopakv: thank you for the quick response. I modified the paylload samples, but running `node bin/octokit-webhooks.js check --cached` did not update `index.json` it seems to fetch from the website (or i might be very wrong).

roopakv: oh nvm i think it was `update` instead of `check`.

roopakv:
@gr2m updated please take a look when you get a chance :)

gr2m (approved):
Looks good, thanks a lot!

PR #180, opened 2020-10-05
feat: always have examples with assignees

christophehurpeau:
Related to https://github.com/octokit/webhooks.js/issues/293

This ensure all "assignees" array are typed


gr2m (approved):
Awesome, thanks a lot!

gr2m (changes requested):
sorry it seams that `index.json` is out of sync with the examples. You can run `node bin/octokit-webhooks.js update --cached` to spot the differences. Could you please have a look?

christophehurpeau:
@gr2m yes thanks for the command ! I think it would be great to add it in CONTRIBUTING.md :)
However I don't quite understand why there is some some `"assignees": []` back ? is there generation from something ouside payload-examples ?

christophehurpeau:
Is that ok if I make a PR for each field or should I do one big PR ?

gr2m:
> I think it would be great to add it in CONTRIBUTING.md :)

Very good idea 👍🏼 

> Is that ok if I make a PR for each field or should I do one big PR ?

Of course! I'll merge this PR in, and then you can do follow up PRs as you see fit

gr2m (approved):

PR #182, opened 2020-10-05
feat: always have examples with labels

christophehurpeau:
Related to octokit/webhooks.js#293

This ensure all "labels" array are typed

### DIFF
--- a/index.json
+++ b/index.json
@@ -6834,7 +6834,16 @@
             "type": "User",
             "site_admin": false
           },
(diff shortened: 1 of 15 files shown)


gr2m (approved):
Thanks so much Christophe, these improvements will make lots of folks very happy, especially once we update the types for https://github.com/octokit/webhooks.js. That will happen tonight or tomorrow via a the nightly Dependabot update

PR #183, opened 2020-10-05
feat: always have examples with requested_reviewers

christophehurpeau:
Related to octokit/webhooks.js#293

This ensure all "requested_reviewers" array are typed

It however doesn't make sense that the creator is the same as the requested_reviewer, is there another example user I could use ?


christophehurpeau:
Oups, I'll fix the conflicts

gr2m:
> It however doesn't make sense that the creator is the same as the requested_reviewer, is there another example user I could use ?

Maybe https://api.github.com/users/octocat?

christophehurpeau:
@gr2m yes thanks ! updated with that user :)

gr2m (approved):
Thanks a lot!

PR #184, opened 2020-10-07
fix(payload-examples): Add missing "organization" object

wolfy1339:
Adds the missing organization object in multiple events.
Ref: octokit/webhooks.js#87


dominguezcelada (approved):

gr2m (changes requested):
Unfortunately, we cannot just add `organization` to the existing payloads, because it woudl imply that it's always present which is not the case. 

> Webhook payloads contain the organization object when the webhook is configured for an organization or the event occurs from activity in a repository owned by an organization.

https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#webhook-payload-object-common-properties

Instead, we have to create additional payload examples, e.g. copy `payload-examples/api.github.com/check_run.completed.payload.json` into sth like `payload-examples/api.github.com/check_run.completed.with-organization.payload.json`. Because the property will be present in one payload but not the other, the resulting types will make the `organization` property optional

wolfy1339:
Got it :+1: . Will do that and re-push

wolfy1339:
@gr2m I can't seem to update the `index.json` file anymore, it errors out.
```
Error: https://github.com/github/docs/edit/main/content/developers/webhooks-and-events/webhook-events-and-payloads.md has no #anchor hash
```

dominguezcelada (approved):

gr2m:
> @gr2m I can't seem to update the `index.json` file anymore, it errors out.
> 
> ```
> Error: https://github.com/github/docs/edit/main/content/developers/webhooks-and-events/webhook-events-and-payloads.md has no #anchor hash
> ```

I just fixed that a moment ago, via #185

gr2m (approved):
Looks good to me 👍🏼 Maybe it would make sense to rephrase the commit message to something like

> feat(examples): event payloads with organization key

I'll leave it up to you, ship at your convenience :)

PR #195, opened 2020-10-13
fix: add event examples with installation key that don't have them yet

### LINKED ISSUE #198: Add support for optional parameter `installation` for Webhook Payloads missing it
Context here: https://github.com/octokit/webhooks.js/issues/266

dominguezcelada:
## 📝 Summary

Add `installation` optional paramater to Webhook Payloads missing it.

## ⛱ Motivation and Context

https://github.com/octokit/webhooks.js/issues/266
Fix #198


wolfy1339 (approved):

payload-examples/api.github.com/installation_repositories.added.payload.without-installation-prop.json
```diff
@@ -0,0 +1,34 @@
...
+    "starred_url": "https://api.github.com/users/Codertocat/starred{/owner}{/repo}",
+    "subscriptions_url": "https://api.github.com/users/Codertocat/subscriptions",
+    "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+    "repos_url": "https://api.github.com/users/Codertocat/repos",
+    "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+    "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+    "type": "User",
+    "site_admin": false
+  }
+}
```
gr2m, line 34: The `installation_repositories` event has always the `installation` key.

payload-examples/api.github.com/installation.created.payload.without-installation-prop.json
```diff
@@ -0,0 +1,32 @@
...
+    "starred_url": "https://api.github.com/users/Codertocat/starred{/owner}{/repo}",
+    "subscriptions_url": "https://api.github.com/users/Codertocat/subscriptions",
+    "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+    "repos_url": "https://api.github.com/users/Codertocat/repos",
+    "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+    "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+    "type": "User",
+    "site_admin": false
+  }
+}
```
gr2m, line 32: The `installation` event has always an installation key

gr2m (changes requested):
remove the example payloads for `installation` and `installation_repositories` that don't have the `installation` key. Also the installation always has a `node_id` which is a string besides the `id` key

payload-examples/api.github.com/installation.created.payload.without-installation-prop.json
dominguezcelada: I doubted about this one because, according to docs, it says
> The GitHub App installation. Webhook payloads contain the installation property when the event is configured for and sent to a GitHub App.

https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#installation

payload-examples/api.github.com/installation_repositories.added.payload.without-installation-prop.json
dominguezcelada: I doubted about this one because, according to docs, it says
> The GitHub App installation. Webhook payloads contain the installation property when the event is configured for and sent to a GitHub App.

https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#installation_repositories

gr2m: go PR the docs, I'm very sure that `"installation"` is always present for these two particular events. They will let us know if otherwise. You can reference this PR

payload-examples/api.github.com/installation.created.payload.without-installation-prop.json
gr2m: In this case the `"installation"` object is the key part of the payload, it has many properties beyond just `id` and `node_id`. I'm sure that it's always present

dominguezcelada:
Removed `installation` and `installation_repositories` examples without `'installation'` property until https://github.com/github/docs/issues/516 is clarified to not block https://github.com/octokit/webhooks.js/issues/266

gr2m (approved):

gr2m:
Great work!

dominguezcelada:
> Removed `installation` and `installation_repositories` examples without `'installation'` property until [github/docs#516](https://github.com/github/docs/issues/516) is clarified to not block [octokit/webhooks.js#266](https://github.com/octokit/webhooks.js/issues/266)
> 
> Once Github Docs provides feedback, a separate PR would be opened in case it's necessary to tackle `installation?` prop.

Confirmed by Github Docs team. The property is there 100% so no need to do a followup PR :)

PR #197, opened 2020-10-13
issue-templates: fix folder name to make it work on Github

dominguezcelada:
Fix #197


wolfy1339 (approved):

gr2m:
no harm done, but this change did not require a new release. I'd have used the `docs` prefix, or no prefix at all

dominguezcelada:
> no harm done, but this change did not require a new release. I'd have used the `docs` prefix, or no prefix at all. When thinking about the commit message prefixes `fix` and `feat`, always think of the perspective of the consumer of this library. If the change has no effect for them, don't use these prefixes

Understood! I will keep in mind for future commits 🙏🏽

PR #222, opened 2020-12-03
fix: add missing example for check_run.requested_action

codebytere:
This PR adds a missing payload example for [check_run.requested_action](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#check_run). I copied existing approaches from the repo but please let me know if i'm missing anything!

cc @gr2m

### DIFF
--- a/index.json
+++ b/index.json
@@ -1592,6 +1592,318 @@
         }
       },
       {
+        "action": "requested_action",
+        "check_run": {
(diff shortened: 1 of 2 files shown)


gr2m (approved):
This is perfect, thank you Shelley!

PR #225, opened 2020-12-06
feat: add payload for organisation webhook

G-Rath:
Per comment in #3


gr2m (approved):
Thanks Gareth! 

I realized that the naming conventions of the payload example files are significant, and also that `node bin/octokit-webhooks.js update --cached` needs to be run after payload example changes. I mention it in the `CONTRIBUTING.md` file, but I also added `payload-examples/api.github.com/README.md` for instructions specific to that folder.

I've made the necessary updates. The update of the types `@octokit/webhooks` should follow soon via an automated pull request, stay tuned

PR #230, opened 2020-12-12
feat: add examples of payloads with organizations

G-Rath:
I'm sure there's more, but so far I've only had these events :)

Also included adding an example of a GH alert, which has two extra properties.


G-Rath:
@gr2m I *think* I've got the naming convention correct on this one, but happy to be schooled if not :)

G-Rath:
```
No payload examples for secret_scanning_alert
❌  webhooks are not up-to-date
```

iirc last week secret scanning was marked as in beta, so I'm guessing it's now released publically? (at least, from the "select webhook events" interfaces POV)

gr2m:
did you run `node bin/octokit-webhooks.js update --cached`?

G-Rath:
@gr2m I have now! Forgot about that step.

gr2m (approved):

PR #233, opened 2020-12-17
feat: add example of `removed_from_repository` action for team event

G-Rath:
I compared this to `added_to_repository` and confirmed they're the same except for the action, so just copied that for the values.

I think that `repository.permissions` has had more properties added to it since the `added_to_repository` payload example was captured - I suspect that applies to the other payloads too.

I'll probably make a followup PR updating all the examples at some point.


gr2m (approved):
Thanks!

PR #235, opened 2020-12-18
feat: add example of `repo.edit` action payload

G-Rath:
It'd be nice if we could confirm what properties are possible in `changed` - I'm guessing its:

```
name: string
description: string
homepage: string
private: boolean
visibility: string
has_issues: boolean
has_projects: boolean
has_wiki: boolean
is_template: boolean
default_branch: string
allow_squash_merge: boolean
allow_merge_commit: boolean
allow_rebase_merge: boolean
delete_branch_on_merge: boolean
archived: boolean
```

based on the ["update a repository"](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#update-a-repository) api params, but they might have different names.

This is more important for when the schema is made, since otherwise it'll have to use `additionalProperties` (not the end of the world for a first iteration but would be nice at *some point* in the future to nail down all the possible properties).


gr2m:
This is how you could find out:

1. Create a test repository
2. Go to settings -> Webhooks
3. Press <kbd>add webhook</kbd>
4. In 2nd tab, open https://smee.io/new, copy the URL you get redirected to, leave the tab open
5. Paste the URL into Payload URL
6. Content Type: `application/json`
7. Leave secret empty, has no effect
8. "Which events would you like to trigger this webhook?" -> "Let me select individual events" -> Uncheck push, check repositories

Now you can make changes in your repo and see the payloads on smee in your browser. THe events are only stored in your browser, so you must leave the tab open while you are testing

G-Rath:
@gr2m it's more that I'd have to edit everything in the repo to confirm and hope I'd not missed anything 😅

If I have a chance I'll do a quick edit of the properties on my test repo to check the obvious ones, but I think having best-guesses in the schema and subsequent types should be good enough - if we miss any people will hopefully open an issue.

gr2m:
> @gr2m it's more that I'd have to edit everything in the repo to confirm and hope I'd not missed anything 😅

Fair enough :D I though I just use the opportunity to show you the path how you can test / inspect webhooks fairly easily. Now that I think about it, maybe we could build a tool that could receive webhooks and would load the schema from this repository and then validate all webhooks. If any of them does not validate, the tool could create an issue automagically together with the payload ✨ That would be a fun little side project to work on.

Holler if anyone would be interested!

> if we miss any people will hopefully open an issue

Absolutely correct! Even better, if the tool above existed, people could reproduce the problem with the missing schema by setting up a webhook to the provided URL 😁

gr2m:
shall we just merge it as is? Or do you want to add any other examples?

G-Rath:
> I though I just use the opportunity to show you the path how you can test / inspect webhooks fairly easily

Cheers, but I can give you an easier way: the events sent to webhooks are shown on the webhooks settings - in fact, the webhook itself doesn't even need to succeed, so you can create one with a URL pointing to nowhere and be able to capture the payloads & headers:

image

(Not that I'm ungrateful or anything - I didn't know about https://smee.io - it'll be useful for other webhooks that like to verify before sending❤️)

> That would be a fun little side project to work on.

Oh that sounds awesome! I might have a go at that right now (if it's alright with you?)

> shall we just merge it as is? Or do you want to add any other examples?

Let's do it as is :)

gr2m:
> Oh that sounds awesome! I might have a go at that right now (if it's alright with you?)

absolutely!

gr2m (approved):

PR #239, opened 2020-12-24
chore: add prettier config

G-Rath:
This ensures that the default prettier settings are used for this repo - i.e I've got [my default `.prettierrc`](https://github.com/ackama/prettier-config-ackama) in my home directory, which means without this in the repo running prettier uses that.


gr2m:
haw man I really am not a big fan of adding dot files. Any chance there is another workaround for your case?

G-Rath:
You can also provide the config in the `package.json` - usually not a fan of doing that since it increases the package size a little by being in the `package.json`, but in this case it's just an empty object so 🤷 

Also just noticed `prettier` is listed as a dependency when it's only a dev dependency, so will open a PR to fix that :)

gr2m (approved):
That works 👍🏼 Thanks!

PR #241, opened 2020-12-29
fix(deps): move `prettier` to `devDependencies`

G-Rath:
Prettier is not used by this project at "runtime", so it's only needed for deving.


gr2m (approved):
Thanks!

PR #186, opened 2020-10-07
feat: json schemas for payloads

heuels:
Hey @gr2m,

In this PR I've added JSON schemas for every payload in `payload-examples`. In order to run a validation, run the following command:

```sh
bin/octokit-payloads.js
```

Closes #136.

PR #243, opened 2021-01-04
chore: Rename files, so that they are consistent

wolfy1339:
Also fixes some typos in the naming of the files.

No payloads were changed or added in this PR.


gr2m:
will this create conflicts for https://github.com/octokit/webhooks/pull/186? I really want to land https://github.com/octokit/webhooks/pull/186 as soon as possible

wolfy1339:
It should not. It uses `index.json` to create the schemas, and uses `fs.readdirSync()` to get all filenames in the `payload-examples` directory.

gr2m (approved):
Thanks!

PR #247, opened 2021-01-07
feat: Add installation.new_permissions_accepted payload and schema

wolfy1339:
Adds a new example payload for `installation.new_permissions_accepted`
Makes installation.schema.json more explicit on types, by using enums and adding new properties


gr2m (approved):
Great work!

PR #248, opened 2021-01-07
feat: Use the common schemas for more properties of various schemas

wolfy1339:


wolfy1339:
@G-Rath It's complaining that there is no schema for an event, but there is.
Can you take a look?

payload-schemas/schemas/common/repository.schema.json
```diff
@@ -429,6 +429,9 @@
         }
       },
       "additionalProperties": false
+    },
+    "public": {
+      "type": "boolean"
```
gr2m, line 434: I'm pretty sure that the `"public"` key is always present, can you add it to the `"required"` array?

gr2m (commented):
Just a small nit, otherwise looks good to me

payload-schemas/schemas/common/repository.schema.json
wolfy1339: The `"private"` key is always present, but not `"public"`.
I checked the payload for `repository.transferred`, `repository.created` and the `"public"` key is not included in the `"repository"` object.

It seems that it's only used for the `fork` event from what I can tell

gr2m: You are right, sorry, I mixed up the two

gr2m (approved):

PR #251, opened 2021-01-07
fix: installation and organization properties are never null

wolfy1339:
They are either present or not (undefined), and not null.


G-Rath (approved):
I suspected this might be the case

Issue #250, opened 2021-01-07
False-positives for missing schemas when an error occurs during schema validation


wolfy1339:
**What happened?**

Running `node bin/validate-payload-examples.js` with a typo in the path to a common schema throws an error that is then falsely logged as missing schema

**What did you expect to happen?**

For the logged error message to match the actual error thrown in the code

**What the problem might be**
Instead of sending an error message related to the actual error, we swallow the error that was thrown and falsely assume it means that the schema does not exist

https://github.com/octokit/webhooks/blob/master/bin/validate-payload-examples.js#L24-L26

PR #252, opened 2021-01-07
feat: adjust structure of schema & examples for payloads to be unions

G-Rath:
Moves schemas to be structured in folders rather than as a single json schema for each event.
Moves examples to match that structure too.

This gives us the advantage of being able to define the schemas for each *action* an event has, making things more accurate.
When we merge the schemas, we create a union of each possible action for an event which we reference at the root - if an event doesn't have actions, its schema is called `event.schema.json` and is referenced directly to avoid a one-type union (there are still a few one-type unions for events that have actions but currently only one action, but that's fine because in theory the event could get more actions down the line).

image


G-Rath:
@gr2m @wolfy1339 would you be happy if I were to do some refactoring of the js files, and looked to move them to TypeScript to make them easier to work with?

They're not in a super bad state or anything, but I found they were inconsistent in a few places with some checks, and think having type annotations would make things clearer.

wolfy1339:
> @gr2m @wolfy1339 would you be happy if I were to do some refactoring of the js files, and looked to move them to TypeScript to make them easier to work with?
I think having types would be a great addition. The issue of converting scripts to typescript has come up in another repository, and it was deemed un-necessary.
A great compromise would be using `// @ts-check` along with JSDoc to add as much detail and documentation

bin/validate-payload-examples.js
```diff
@@ -6,26 +6,29 @@ const { ajv, validate } = require("../payload-schemas");
...
+          console.error(
+            `❌ Payload '${event}/${filename}' does not match schema`
+          );
+          console.error(ajv.errors);
+          hasErrors = true;
+        } else {
+          console.log(`✅ Payload '${event}/${filename}' matches schema`);
+        }
+      } catch (err) {
+        console.error(`Missing schema for event '${event}'`, err);
```
wolfy1339, line 27: I don't think this change would be the best way to log errors. See #250

bin/octokit-schema.js
```diff
@@ -89,4 +133,4 @@ async function run() {
   }
 }
 
-run();
+run().catch(console.error);
```
wolfy1339, line 136: Can we try to handle errors instead of just logging them?

G-Rath:
Awesome, I'll do that along with writing the `otcokit-type` script for generating the types :D

> The issue of converting scripts to typescript has come up in another repository, and it was deemed un-necessary.

I don't mind using `// @ts-check`, but if we're happy with `.ts` I'd prefer to do it with that as then we get some basic transpiling meaning we can use things like optional chaining without having to do any building since we can just call [`ts-node -T`](https://github.com/TypeStrong/ts-node) as a drop-in replacement for just `node` 🎉

bin/validate-payload-examples.js
G-Rath: Whoops yeah forgot I did that 😅 

I was going to fix that once I'd finished this PR - what do you think should happen, log or throw?

bin/octokit-schema.js
G-Rath: This is here because it's otherwise technically a floating promise - otherwise it'll be an uncaught rejection error. I originally added it to make sure I wasn't missing errors when things seemed to just not be working, so it can be removed.

wolfy1339:
Let's not forget to update [`@octokit/webhooks`](https://github.com/octokit/webhooks.js) with all these changes since we created the schemas

G-Rath:
That'll require the generating of types from the schemas, which is what I'm doing now as it required this PR - then I'll update `@octokit/webhooks` to use those types.

gr2m:
> move them to TypeScript to make them easier to work with?

My main motivation to keep the scripts in JS was to lower the barrier for contributors. But times are changing, more and more folks are comfortable with TS and if you both would prefer the scripts to be in TS, I don't mind the change. You've been a great help, let's do whatever makes it easier / more fun to you

G-Rath:
@wolfy1339 I've rebased with the changes you made in #256 :)

wolfy1339 (approved):
This is great. It will allow us a lot of more explicit types.
This looks good to me. :pray:

G-Rath:
Awesome - shall we merge, or wait for @gr2m to approve?

wolfy1339:
Lets wait

bin/octokit-schema.js
```diff
@@ -4,25 +4,37 @@ const path = require("path");
...
-      $ref: `payload-schemas/schemas/common/${schema}`,
+      $ref: `${pathToWebhookSchemas}/common/${schema}`,
     };
   });
 
   return definitions;
 };
 
+const listEvents = () => {
+  const dirents = fs.readdirSync(pathToWebhookSchemas, {
```
gr2m, line 29: what is a dirent? Maybe call the variable `files`? Or something more descriptive? I prefer to avoid any kind of abbreviations to make the code easier to read / maintain / contribute to

gr2m (approved):
This is great 👍🏼 Just one minor nit regarding variable naming, but otherwise good to merge

bin/octokit-schema.js
G-Rath: It's a [directory entity](https://nodejs.org/api/fs.html#fs_class_fs_dirent), which is what `readdirSync` returns when `withFileTypes` is `true`.

I agree with avoiding abbreviations, but in this case it's the name of the actual underlying class and the expanded version is very long `directoryEntity` - the reason I don't want to call it something like `files` is that it's not a file: it has methods on it that you can call.

gr2m: `directoryEntity` is not long, you can call it `directoryEntities` or `directoryEntries`. I was just confused by `dirents`, I assume others will, too

G-Rath: sorry your right here it'd be called `directoryEntities` - but does that mean you're happy for `dirent` to be used for the `filter` & `map` parameters?

gr2m: I'd call it `directoryEntity` in both, or `entity` would also work in this context, better than `dirent`, I think

G-Rath:
> otherwise good to merge

Do you want to hit the button, or should I? 😁

PR #254, opened 2021-01-08
Convert scripts to typescript

G-Rath:
This is branched off #252 to avoid me having to do an ugly rebase - ideally that should be merged first to make this review-able

----

What started as converting the scripts to TS ballooned slightly, so this includes a few other maintenance improvements:

* moving all the scripts to TypeScript - this uncovered a few small nits, and also revealed that `getSections` doesn't need to be async which is cool 🎉 
* replacing `fs-extra` with native methods - most of the methods were passthroughs from `fs`, and the read & write for `JSON` wasn't being used
  * I did replace the use of `exists` with a `try/catch` - fs existence checks are usually not recommended as they're technically a race condition, so the throw could always happen
  * tbh I think its probably worth just inlining `cache` into `get-html` as thats the only place it's used
* removed `execa` dev dependency as its unused
* updating the workflows to all use the same node & action versions (node 12)
* ran [`sort-package-json`](https://www.npmjs.com/package/sort-package-json) for consistency (happy to revert this if people want - I just find it nice to not have to worry about the order being consistent)


README.md
```diff
@@ -256,7 +256,7 @@ import SCHEMA from "@octokit/webhooks-definitions/schema.json";
 
 ## How it works
 
-This package updates itself using a daily cronjob running on GitHub Actions. The [`index.json`](index.json) file is generated by [`bin/octokit-webhooks.js`](bin/octokit-webhooks.js). Run `node bin/octokit-webhooks --usage` for instructions. After the update is complete, run `npm run build:webhooks` and `npm run build:schema` to update `index.json` and `schema.json` files.
+This package updates itself using a daily cronjob running on GitHub Actions. The [`index.json`](index.json) file is generated by [`bin/octokit-webhooks.ts`](bin/octokit-webhooks.ts). Run `node bin/octokit-webhooks --usage` for instructions. After the update is complete, run `npm run build:webhooks` and `npm run build:schema` to update `index.json` and `schema.json` files.
```
wolfy1339, line 259: Looks like you forgot a `ts-node` here

lib/types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface Section {
+  id: string;
+  html: string;
+}
+
+// polyfills for types used by turndown to avoid having the include the DOM lib
+// as they'd be considered globally available in the whole codebase by typescript
+// (remove if we actually start using DOM elements e.g. via jsdom)
+declare global {
```
wolfy1339, line 21: Could you explain the reasoning here? Why wouldn't we just include the DOM lib, and call it a day

G-Rath: I try to avoid including the DOM lib on non-dom projects because you can't scope their availability to specific files easily - so it becomes "valid" to use them anywhere, and autocompletion in IDEs will start drawing from those types which can be confusing.

It's also not easy to tell when reviewing if a type is coming from the DOM, e.g. `Element` is a type in the DOM lib, meaning you could reference it by mistake when you really want `cheerio.Element`. (a few other good ones include `Url` , `Blob`, and `Timer` - that's an especially annoying one as `NodeJS.Timer` doesn't have the same type as `DOM.Timer`)

Hence for situations like this (where we're just doing `turnabout.turnabout`, and so not interacting with the section that uses the DOM libs) I like to favor declaring empty interfaces so I don't have to worry about it.

README.md
G-Rath: 🤦 good catch!

gr2m:
> * moving all the scripts to TypeScript - this uncovered a few small nits, and also revealed that `getSections` doesn't need to be async which is cool 🎉
> 
> * replacing `fs-extra` with native methods - most of the methods were passthroughs from `fs`, and the read & write for `JSON` wasn't being used
>   
>   * I did replace the use of `exists` with a `try/catch` - fs existence checks are usually not recommended as they're technically a race condition, so the throw could always happen
>   * tbh I think its probably worth just inlining `cache` into `get-html` as thats the only place it's used
> 
> * removed `execa` dev dependency as its unused
> 
> * updating the workflows to all use the same node & action versions (node 12)
> 
> * ran [`sort-package-json`](https://www.npmjs.com/package/sort-package-json) for consistency (happy to revert this if people want - I just find it nice to not have to worry about the order being consistent)

these all sound like great changes, thank you! I'm looking through it now. But I'd like @wolfy1339 and @oscard0m to give their thumbs up for this change before we move forward

G-Rath:
No problem! I can break a few of these changes out into another PR if you'd like (specifically, removing `execa`, updating the workflows, and applying `sort-package-json`) :)

CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ If you figure out how to fix the issue then please push the fix to the pull requ
 The [`index.json` file](https://github.com/octokit/webhooks/blob/master/index.json) is generated, please do not edit it. Instead, make changes in the [`payload-examples/api.github.com/` folder](https://github.com/octokit/webhooks/tree/master/payload-examples/api.github.com), then update `index.json` by running the following command
 
 ```
-node bin/octokit-webhooks.js update --cached
+ts-node -T bin/octokit-webhooks.ts update --cached
```
gr2m, line 20: Folks won't have the `ts-node` binary installed globally and might not be familiar with it. I'd run the command with `npx` which will use the installed version from our `devDependencies`

```suggestion
npx ts-node -T bin/octokit-webhooks.ts update --cached
```

gr2m: alternatively, we could add an npm script?

README.md
```diff
@@ -256,7 +256,7 @@ import SCHEMA from "@octokit/webhooks-definitions/schema.json";
 
 ## How it works
 
-This package updates itself using a daily cronjob running on GitHub Actions. The [`index.json`](index.json) file is generated by [`bin/octokit-webhooks.js`](bin/octokit-webhooks.js). Run `node bin/octokit-webhooks --usage` for instructions. After the update is complete, run `npm run build:webhooks` and `npm run build:schema` to update `index.json` and `schema.json` files.
+This package updates itself using a daily cronjob running on GitHub Actions. The [`index.json`](index.json) file is generated by [`bin/octokit-webhooks.ts`](bin/octokit-webhooks.ts). Run `ts-node -T bin/octokit-webhooks --usage` for instructions. After the update is complete, run `npm run build:webhooks` and `npm run build:schema` to update `index.json` and `schema.json` files.
```
gr2m, line 259: Run `ts-node` with `npx`. 

```suggestion
This package updates itself using a daily cronjob running on GitHub Actions. The [`index.json`](index.json) file is generated by [`bin/octokit-webhooks.ts`](bin/octokit-webhooks.ts). Run `npx ts-node -T bin/octokit-webhooks --usage` for instructions. After the update is complete, run `npm run build:webhooks` and `npm run build:schema` to update `index.json` and `schema.json` files.
```

or alternatively let's add an npm script

gr2m (commented):
just a few comments from me, thanks for all your help!

README.md
G-Rath: I'm keen for npm scripts!

CONTRIBUTING.md
wolfy1339: There actually was one introduced in #186.
```suggestion
npm run build:webhooks
```

gr2m: looks good to me 👍🏼

G-Rath: Have updated it to be `npm run build:webhooks -- --cached` :)

README.md
G-Rath: Have replaced it with `npm run octokit-webhooks -- --usage` :)

wolfy1339 (approved):

oscard0m (approved):
> * moving all the scripts to TypeScript - this uncovered a few small nits, and also revealed that `getSections` doesn't need to be async which is cool 🎉

Amazing Job! Thanks for your time and knowledge! It has been a pleasure reading your TS part 🤓 

> * replacing `fs-extra` with native methods - most of the methods were passthroughs from `fs`, and the read & write for `JSON` wasn't being used
>   
>   * I did replace the use of `exists` with a `try/catch` - fs existence checks are usually not recommended as they're technically a race condition, so the throw could always happen

Didn't know about this. Adding a link just in case somebody want to take a look into it> https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

>   * tbh I think its probably worth just inlining `cache` into `get-html` as thats the only place it's used

Yep, to me has sense to unify files :)

> * removed `execa` dev dependency as its unused

✅ 

> * updating the workflows to all use the same node & action versions (node 12)

🥳 🍾 

> * ran [`sort-package-json`](https://www.npmjs.com/package/sort-package-json) for consistency (happy to revert this if people want - I just find it nice to not have to worry about the order being consistent)

✅ To me it's fine. Maybe we can open an issue to automate this step. What do you think? @G-Rath @gr2m @wolfy1339 I found this Prettier plugin for instance https://github.com/matzkoh/prettier-plugin-packagejson

---

💬 Left a comment regarding tsconfig.json. If you want to move forward and open a different issue to tackle that to me it's completely fine.

---

A general feedback:
- 🥇 Amazing all the job you did here and other PR's. Congratulations and thanks for your effort and for sharing your knowledge with the community.

- 💬 Personally as a contributor and as a reviewer I would prefer to get more PR's with a single purpose or change (just a personal preference 😃). It's easier to read and review, easy to categorize when using semantic versioning and, in case of a regression, easier to revert too. What are your thoughts on this @G-Rath ?

G-Rath:
@oscard0m thanks for such an awesome write-up!

> I would prefer to get more PR's with a single purpose or change

Couldn't agree more - that ones totally on me! Usually I would've done it like that, but all the rebasing and creation/deletion of so many files very quickly tends to upset my file system (specifically the WSL-based watchers) 😬 

> Amazing all the job you did here and other PR's

No problem - thanks for being so engaging, it makes the contributions feel a lot more valuable. Feel free to ping me elsewhere if you'd like a hand or have questions, especially with Typescript-related stuff 😄 (that goes for @wolfy1339 & @gr2m too - you've all been really fun to work with!)

> Maybe we can open an issue to automate this step

Can do! The prettier plugin you linked immediately came to mind - I'll set it up in a follow up PR.

------

Since this PR is branched off #252, I'll apply the changes we've discussed to this branch while that PR is still open, and then will open new PRs for whatever ones I don't get around to addressing when its merged :)

G-Rath:
@gr2m @oscard0m I didn't realise that squashing was the merge method of choice for this repo so doubly sorry about not doing a pr for each commit 😱

Are you happy to merge via rebase? otherwise I'll do each commit via their own PR

gr2m:
no worries, PRs can be as messy as it makes sense. We can squash all the commits into a single one at the time of the merge, no need for rebases

G-Rath:
I was meaning using "Rebase and Merge" rather than "Squash and Merge" into the base branch, since these commits are pretty unrelated to each other and so squashing them into a single commit would... well have them as a single big commit 😅

gr2m:
> well have them as a single big commit 😅

that's all right, we still have the separate commits in the PR. Don't worry too much about it. If you like, we can do a rebase with cleaned up commits, but I think it's not the most fun thing to do ;) we'll be fine

G-Rath:
Since #252 has been merged, this PR can be landed - I didn't get around to addressing the extras, so I'll do them in follow-up PRs.

PR #255, opened 2021-01-08
feat: generate types from schema

G-Rath:
This is branched off #252 to avoid me having to do an ugly rebase - ideally that should be merged first to make this review-able

-----

Look, it's type generation!

Also ported over the code to generate the `EventPayloadMap` interface that can be used to do dynamic mapping.

~I've made this a draft for now as while I think it's ready for review, it requires #252 and I think we should discuss how we're handling the build flow - I think it could benefit from a bit of a revision now that we've got a bit of a chain going.~

~In particular, do we want to check-in the `schema.d.ts`? Personally I think it would make reviewing easier, but we currently don't check-in the `schema.json` which is required by this script.~

~(this is currently written in TS, but can convert to JS easily if it looks like #254 isn't going to be merged for a bit)~


package.json
```diff
@@ -33,13 +36,15 @@
     "@types/cheerio": "^0.22.23",
     "@types/json-diff": "^0.5.0",
     "@types/json-schema": "^7.0.6",
+    "@types/lodash": "^4.14.167",
```
G-Rath, line 39: This is needed as a fix until https://github.com/bcherny/json-schema-to-typescript/pull/362 is landed.

G-Rath:
can't statically import a file that's generated by a build command 🤦

package.json
```diff
@@ -10,15 +10,19 @@
...
   "scripts": {
     "build": "ts-node -T bin/octokit-webhooks.ts check --cached",
+    "build:all": "npm run -s build:webhooks && npm run -s build:schema && npm run -s build:types",
     "build:schema": "ts-node -T bin/octokit-schema.ts",
+    "build:types": "ts-node -T bin/octokit-types.ts",
     "build:webhooks": "ts-node -T bin/octokit-webhooks.ts update",
     "lint": "prettier --check '{bin,lib}/*.ts' '*.{ts,md}' '{payload-examples,payload-schemas}/**/*.json' 'payload-schemas/*.ts' *.md package.json index.json",
     "lint:fix": "prettier --write '{bin,lib}/*.ts' '*.{ts,md}' '{payload-examples,payload-schemas}/**/*.json' 'payload-schemas/*.ts' *.md package.json index.json",
     "octokit-schema": "ts-node -T bin/octokit-schema.ts",
+    "octokit-types": "ts-node -T bin/octokit-types.ts",
```
gr2m, line 25: Can we remove one of the two scripts running `ts-node -T bin/octokit-types.ts`?

gr2m (approved):
I think eventually I'd move the event types into a different repository, because they are only relevant to the JavaScript Octokit projects, but for now we can add it here.

wolfy1339 (approved):

gr2m:
Could you add a section to the README on how to import the types? It would be good to show how to import the type for a specific event payload. That would answer @AlCalzone's question at https://github.com/probot/probot/issues/1456

G-Rath:
@gr2m done - I'll merge this shortly, and start on my refactor of webhooks.js & co 🎉

gr2m:
Great work everyone

PR #260, opened 2021-01-11
chore: format payload and schema jsons with prettier consistently

G-Rath:
The backstory is that while `prettier` formats things consistently, with objects you can "choose" if they're on a single line or not based on if you put a newline after the opening brace - while I think this is a great thing, this is a rare case where we don't want that since such newlines are lost when you parse raw json into js objects, meaning noisy diffs when you're working with scripts and especially when you're wanting to compare two versions of schemas to check if the output is what you're wanting or if you've messed something up.

I've made this PR as a once-off reformatting as I've got a bunch of bulk-changes to the schemas to land that are much easier to review when the formatting is consistent - the script for doing will have its own PR.


gr2m (approved):
good to merge, you'll have to resolve the merge conflicts though

PR #261, opened 2021-01-11
chore: add script for formatting schemas & payloads with `prettier`

G-Rath:
This is the script that made #260

The backstory is that while `prettier` formats things consistently, with objects you can "choose" if they're on a single line or not based on if you put a newline after the opening brace - while I think this is a great thing, this is a rare case where we don't want that since such newlines are lost when you parse raw json into js objects, meaning noisy diffs when you're working with scripts and especially when you're wanting to compare two versions of schemas to check if the output is what you're wanting or if you've messed something up.

I'm on the fence about making this script "required", as it takes a few seconds to run since you're reading in every json file formatting and then writing them again, and it's not something I'm expecting to pop up day-to-day - more I expect this being run every so often before doing some change that touches every json file.

Happy to bikeshed over the name, ci/script implementation etc

### DIFF
--- a/bin/format-with-prettier.ts
+++ b/bin/format-with-prettier.ts
@@ -1,0 +1,39 @@
+#!/usr/bin/env ts-node-transpile-only
+
+import fs from "fs";
+import { JSONSchema7 } from "json-schema";
+import { format } from "prettier";
+
+const checkOnly = process.argv.includes("--check");
+
+const formatJsonInDirectory = (pathToJsons: string) => {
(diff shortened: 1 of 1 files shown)


gr2m:
Maybe for now let's just add the script and run it as needed. We can still automate things later once it becomes a pain?

G-Rath:
Yeah, I think that's the approach to go with.

gr2m (approved):

wolfy1339 (approved):

PR #262, opened 2021-01-11
build: make `bin/*` scripts executable

gr2m:
This should fix this CI error https://github.com/octokit/webhooks/runs/1683852412?check_suite_focus=true


wolfy1339 (approved):

G-Rath (approved):
sorry that's my bad - I'm on Windows w/ WSL, and so the file permissions for executables sometimes get clobbered since my IDE is in Windows.

gr2m:
No worries at all

gr2m:
@G-Rath it now fails with 

> Run bin/octokit-webhooks.ts update
/usr/bin/env: ‘ts-node-transpile-only’: No such file or directory
Error: Process completed with exit code 127.

https://github.com/octokit/webhooks/runs/1684467838?check_suite_focus=true

We can fix the CI by running `npm run build:webhooks` instead. I was not able to reproduce the problem locally. Maybe it's Linux/Ubuntu-specific 🤷🏼

G-Rath:
Ah yeah that'll be because it's not installed globally - `env` doesn't know to look into `node_modules` since it's not on the path (it can get tricky locally because most npm-related tools like `nodenv`, `npm`, etc, will push `node_modules/.bin` into the path, but it's not always clear when they do)

Using `npm run <script>` or `npx ts-node -T` will fix it.

PR #264, opened 2021-01-11
feat: various fixes and improvements to schemas

wolfy1339:


gr2m (approved):
Wow these changes look great!

wolfy1339:
@G-Rath If you could review this as well. I've done a lot of changes, just want you to confirm everything will work as intended especially once they get converted to TS types

oscard0m (approved):
This is A-W-E-S-O-M-E! I'm super excited about all the power these changes landing to webhooks. Congrats to everybody 😍

G-Rath:
> @G-Rath If you could review this as well.

Sure thing - I'll try to get to sometime this week (works a bit busy right now, so might have to wait a couple of days)

wolfy1339:
The [openapi spec](https://github.com/github/rest-api-description) for the rest api shares a lot of properties with these shemas.
Later on, we could try and use it as the source of truth where possible.

oscard0m:
> The [openapi spec](https://github.com/github/rest-api-description) for the rest api shares a lot of properties with these shemas.
> Later on, we could try and use it as the source of truth where possible.

Agree on that.

Should we create an issue to follow up this?

wolfy1339:
> Should we create an issue to follow up this?

Yes, for sure.

G-Rath (approved):
Overall looks good.

I'll need to do some checking on how the `enum: [true]` and `allOf`s come out, but I don't think they'll be any real trouble.

There are a few optimizations you can make, but I've started committing my `optimize-schema` script which'll handle them automatically.

I'm happy if you want to merge this :)

wolfy1339:
I'm noticing a lot of events have missing schemas & payload examples for the different actions

oscard0m:
> > Should we create an issue to follow up this?
> 
> Yes, for sure.

https://github.com/octokit/webhooks/pull/264

oscard0m:
> I'm noticing a lot of events have missing schemas & payload examples for the different actions

Should we create an issue for this? Would have sense to tackle it in https://github.com/octokit/webhooks/issues/266?

PR #265, opened 2021-01-13
Create schema optimizer

G-Rath:
This is a pre-commit mainly to get the script created so that I can make PRs adding each optimization one-by-one without as much conflicting.

Like `format-with-prettier`, this isn't something I'm expecting to have to run much after the initial bulk run, but think its worth having committed.

In the future it would be nice to have a tool checking new schemas for the situations this script is handling, but we can do that later.

-----

This first optimization replaces all `anyOf`s with `oneOf`.

`anyOf` means "this should match at least one", while `oneOf` means "this should match at least one, but no more than one".
Situations where you actually want `anyOf` are very rare, as they usually indicate un-normalized schemas since if `d` could match `x` and `y` at the same time that usually means one is a subset of the other (in which case you can just match the super type), or you've got at least two types that are equivalent and so one is duplicating the other.

(I actually struggle to think of a good usecase for `anyOf`, as they all end up with some form of redundant or duplicated parts in the schema).

### DIFF
--- a/bin/optimize-schemas.ts
+++ b/bin/optimize-schemas.ts
@@ -1,0 +1,40 @@
+#!/usr/bin/env ts-node-transpile-only
+
+import { strict as assert } from "assert";
+import fs from "fs";
+import { JSONSchema7 } from "json-schema";
+import { format } from "prettier";
+
+const payloads = "payload-schemas/schemas";
+
+const isJsonSchemaObject = (object: unknown): object is JSONSchema7 =>
+  typeof object === "object" && object !== null && !Array.isArray(object);
+
+fs.readdirSync(payloads).forEach((event) => {
+  fs.readdirSync(`${payloads}/${event}`).forEach((schema) => {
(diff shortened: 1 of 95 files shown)


bin/optimize-schemas.ts
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env ts-node-transpile-only
+
+import { strict as assert } from "assert";
+import fs from "fs";
+import { JSONSchema7 } from "json-schema";
```
wolfy1339, line 5: just a nitpick, but maybe we should start using `import type` when importing types wherever possible

G-Rath: Sure - theres no particular advantage to them in this sort of context, but no harm in it.

oscard0m (approved):

G-Rath:
@wolfy1339 if you approve, I can merge this in and start landing more optimizations :D

wolfy1339 (approved):

PR #267, opened 2021-01-14
fix: use `additionalProperties` wherever we define an object

wolfy1339:
This avoids having `[key: string]: unknown` in the generated types
This reverts part of https://github.com/octokit/webhooks/commit/124b0b284c4b89eac2d8ac0feadee07c7fa43210


G-Rath (approved):
oh this is a great one!

I've chucked this into my optimizer script 😍

wolfy1339:
This has brought up a little problem, you can't use `allOf` to *`extend`* a schema like I was trying to do. Since the second schema object has `"additionalProperties": false`, it will fail.
Another problem is that the current latest version of the json-schema specification (draft-08) is not supported by ajv, which includes some new features that could've helped make this happen.

Now, OpenAPI (which GitHub uses for the REST API) seems to better suit the goal and would make eventually upstreaming the schemas  easier.

wolfy1339:
I seem to have come across a limitation of the current validation system, it's pushing all the schemas for an event into a `oneOf`.
Somehow, it managed to pick the wrong schema and say that there's a missing property, when the `action` property is different from any others from that event., and the property that it's complaining about doesn't exist on the proper schema
I have absolutely no clue as to what I did, as I didn't touch the `required` schema key, nor the `action` property

G-Rath:
> you can't use allOf to extend a schema ... since the second schema object has "additionalProperties": false

Yeah that's right, I remember that being a thing. However, I'm not sure if it's a lot of gain especially as it won't play nicely with the type generator.

> the current latest version of the json-schema specification (draft-08) is not supported by ajv

My understanding is that draft-8 has been renamed to draft-2019-09, which [is supported by ajv](https://github.com/ajv-validator/ajv/blob/master/docs/json-schema.md#json-schema-draft-2019-09)

------

I think this is definitely worth doing, as the advantages of having the objects "sealed" out way what we save by deduplicating lines of code, and there are ways we can improve this in the future.

wolfy1339:
There's probably other ones, but I'd want to wait for your script to do it for me :P

PR #274, opened 2021-01-14
refactor: merge `null` unions in schemas

G-Rath:
Branched off #273

Null can be merged into as a union if its the one other type in the `oneOf`.

### DIFF
--- a/bin/optimize-schemas.ts
+++ b/bin/optimize-schemas.ts
@@ -2,7 +2,11 @@
 
 import { strict as assert } from "assert";
 import fs from "fs";
-import { JSONSchema7, JSONSchema7TypeName } from "json-schema";
+import {
+  JSONSchema7,
+  JSONSchema7Definition,
+  JSONSchema7TypeName,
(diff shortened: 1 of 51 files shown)


wolfy1339:
Is there a particular reason as to why not leave it as-is? I find it's easier to understand, even though it is a little more verbose
How will this prevent problems like this one: https://github.com/G-Rath/compare-gh-webhook-to-schema-function/issues/9

G-Rath:
Because it builds a better union:

image

It also makes it a little easier to move to OpenAPI if desired in the future since you don't have to check as much objects to make sure you've removed the `null`.

G-Rath:
@wolfy1339 https://github.com/G-Rath/compare-gh-webhook-to-schema-function/issues/9 was becuase the type was at the root of the schemas, which means `type` is treated a little differently.

(I actually think it's probably a bug in `json-schema-to-typescript`, which I'll have a look at fixing at some point)

wolfy1339 (approved):

PR #276, opened 2021-01-14
chore: improve validation script

### LINKED ISSUE #250: False-positives for missing schemas when an error occurs during schema validation

G-Rath:
Fixes #250

I figured there's some use in either behaviour depending on what the actual error is, and defaulted to exiting at the first error as that'll make it less explody when you're not expecting an error (since i.e it could be a strict-mode error in one of the common schemas, so every schema that uses it will throw).

### DIFF
--- a/bin/validate-payload-examples.ts
+++ b/bin/validate-payload-examples.ts
@@ -3,7 +3,6 @@
 import fs from "fs";
 import { ajv, validate } from "../payload-schemas";
 
-let hasErrors = false as boolean;
 const payloads = `./payload-examples/api.github.com`;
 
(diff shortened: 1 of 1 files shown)


wolfy1339 (approved):
This is much better!

PR #279, opened 2021-01-14
fix: ensure all `object`s have `additionalProperties` in schemas

G-Rath:
@wolfy1339 the one you've been waiting for 🎉 

Pretty sure this is the last optimization, so next I'll commit my "common schema ref replacer" script, and start creating new common schemas 🥳

### DIFF
--- a/payload-schemas/schemas/label/edited.schema.json
+++ b/payload-schemas/schemas/label/edited.schema.json
@@ -19,7 +19,11 @@
       },
       "additionalProperties": false
     },
-    "changes": { "type": "object", "properties": {} },
+    "changes": {
+      "type": "object",
+      "properties": {},
+      "additionalProperties": false
+    },
     "repository": { "$ref": "common/repository.schema.json" },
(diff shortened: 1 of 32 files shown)


wolfy1339 (approved):

payload-schemas/schemas/label/edited.schema.json
```diff
@@ -19,7 +19,11 @@
       },
       "additionalProperties": false
     },
-    "changes": { "type": "object", "properties": {} },
+    "changes": {
+      "type": "object",
+      "properties": {},
+      "additionalProperties": false
+    },
```
wolfy1339, line 26:
```suggestion
"changes": {
      "type": "object",
      "properties": {
        "color": {
          "type": "object",
          "required": ["from"],
          "properties": { "from": { "type":"string" } },
          "additionalProperties": false
        },
        "name": {
          "type": "object",
          "required": ["from"],
          "properties": { "from": { "type":"string" } },
          "additionalProperties": false
        }
      },
      "additionalProperties": false
    },
```

PR #281, opened 2021-01-15
fix: add schemas for missing event actions

G-Rath:
Noticed these while refactoring `webhooks.js` to use the types. Not sure if this was my restructuring script messing up, or if they've just always been like this 🤷


wolfy1339:
I went looking through the code history, and it seems that there was never an `enum` for the `action` property for these events

wolfy1339 (approved):

PR #287, opened 2021-01-16
fix: add types for items in arrays

wolfy1339:
This adds types to array items where the type is known

### DIFF
--- a/payload-schemas/schemas/installation_repositories/added.schema.json
+++ b/payload-schemas/schemas/installation_repositories/added.schema.json
@@ -111,7 +111,7 @@
           },
           "additionalProperties": false
         },
-        "events": { "type": "array", "items": {} },
+        "events": { "type": "array", "items": { "type": "string" } },
         "created_at": { "type": "integer" },
         "updated_at": { "type": "integer" },
         "single_file_name": { "type": ["string", "null"] }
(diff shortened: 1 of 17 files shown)


payload-schemas/schemas/installation_repositories/added.schema.json
```diff
@@ -134,7 +134,22 @@
...
+          "id": { "type": "integer" },
+          "node_id": { "type": "string" },
+          "name": { "type": "string" },
+          "full_name": { "type": "string" },
+          "private": { "type": "boolean" }
+        },
+        "additionalProperties": false
+      },
+      "maxItems": 0
+    },
```
wolfy1339, line 152: In theory here, this would the it's proper type, however as this is the added schema, this array should be empty, which is why `maxItems` is set to 0

payload-schemas/schemas/installation_repositories/removed.schema.json
```diff
@@ -120,6 +120,22 @@
...
+          "id": { "type": "integer" },
+          "node_id": { "type": "string" },
+          "name": { "type": "string" },
+          "full_name": { "type": "string" },
+          "private": { "type": "boolean" }
+        },
+        "additionalProperties": false
+      },
+      "maxItems": 0
+    },
```
wolfy1339, line 137: In theory here, this would the it's proper type, however as this is the removed action schema, this array should be empty, which is why `maxItems` is set to 0

payload-schemas/schemas/integration_installation_repositories/added.schema.json
```diff
@@ -138,7 +138,22 @@
...
+          "id": { "type": "integer" },
+          "node_id": { "type": "string" },
+          "name": { "type": "string" },
+          "full_name": { "type": "string" },
+          "private": { "type": "boolean" }
+        },
+        "additionalProperties": false
+      },
+      "maxItems": 0
+    },
```
wolfy1339, line 156: In theory here, this would the it's proper type, however as this is the added action schema, this array should be empty, which is why `maxItems` is set to 0

G-Rath (approved):

PR #288, opened 2021-01-16
feat: add teams schema

wolfy1339:


G-Rath (approved):
could you run the schema optimizor on this? you've got a few `anyOf`s :)

Otherwise looks good!

wolfy1339:
The `anyOf` are intetional. The array can contain both or just one

wolfy1339:
The reasoning behind the `anyOf` is that the type can be `[]` an empty array, `Team[]`, `User[]` or an array of both `(Team | User)[]`

G-Rath:
You don't need `anyOf` for that - while JSONSchema looks OO but it's actually not; each item in the data is validated on their own against the `items` schema. Meanwhile we don't want to accept an object that is valid against *both* users and teams.

<details>
<summary> Here's an example</summary>

```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": [
    "object"
  ],
  "definitions": {
    "one": {
      "properties": {
        "name": {
          "type": "string"
        }
      },
      "required": [
        "name"
      ]
    },
    "two": {
      "properties": {
        "age": {
          "type": "number"
        }
      },
      "required": [
        "age"
      ]
    }
  },
  "properties": {
    "arr": {
      "items": {
        "oneOf": [
          {
            "$ref": "#/definitions/one"
          },
          {
            "$ref": "#/definitions/two"
          }
        ]
      }
    }
  },
  "additionalProperties": false
}
```

For the above, this is invalid:

```
{
  "arr": [
    {
      "name": "john"
    },
    {
      "age": 5
    },
    {
      "name": "john",
      "age": 5
    }
  ]
}
```

If you change the `oneOf` to an `anyOf`, it'll be valid :)
</details>

Using `oneOf` will result in the same type as you'd get if you used `anyOf`.

G-Rath (approved):

PR #293, opened 2021-01-16
feat: use common app schema

wolfy1339:
Missed this one


G-Rath (approved):
Have adjust `ref-common-schemas` to handle this in future: #295

PR #294, opened 2021-01-16
fix: adjust required properties

wolfy1339:
Remove properties from `required` that are optional, and removes some all together when that property isn't present


G-Rath (approved):
nice!

PR #298, opened 2021-01-16
feat: create `Milestone` common schema

G-Rath:
Technically this changes the schema for a bunch of schemas, but I'm 99% sure that all instances of milestone can `null` for `closed_at`, and that our payload examples don't include ones with closed milestones for all the schemas that didn't have it marked as `string | null` (about half or so).


wolfy1339:
It seems that you are missing `common/milestone.schema.json`

G-Rath:
Yeah, I keep forgetting to `git add` the extracted schemas 😅

wolfy1339:
> I'm 99% sure that all instances of milestone can `null` for `closed_at`, and that our payload examples don't include ones with closed milestones for all the schemas that didn't have it marked as `string | null` (about half or so).

If you look at the `milestone` event, you'll see some with the `closed_at` property as a string. Those ones, should stay as-is IMO, better to be explicit than to save lines of code

G-Rath:
> If you look at the milestone event, you'll see some with the closed_at property as a string.

I think the only one that applies to is the `closed` event - I'll revert that one shortly.

(I've actually got a cunning plan anyway on how to do `extends` in the types 😉)

oscard0m (approved):

wolfy1339:
> > If you look at the milestone event, you'll see some with the closed_at property as a string.
> 
> I think the only one that applies to is the `closed` event - I'll revert that one shortly.
> 
> (I've actually got a cunning plan anyway on how to do `extends` in the types wink)

Have you gotten around to this?

G-Rath:
> Have you gotten around to this?

No yet sorry 😓

G-Rath:
I'm going to close this due to the rebase conflicts and I think it'll be easier to start again.

Someone else can pick this up if they like, otherwise I'll get back to it eventually :)

PR #300, opened 2021-01-17
fix: installation actions

wolfy1339:
suspended -> suspend
unsuspended -> unsuspend


G-Rath (approved):
Related to https://github.com/octokit/webhooks.js/pull/406

PR #303, opened 2021-01-17
feat: common label schema

wolfy1339:
Ready to merge! 🚀


wolfy1339:
@G-Rath it seems the script missed some places where it could have used `common/label.schema.json`. I had to add them manually in 5448740

G-Rath:
@wolfy1339 it's because those schemas are different: the common one *required* some properties, whereas the ones you've replaced manually don't require those properties :)

wolfy1339:
I have confirmed the changes with a recent payload of mine.

G-Rath (approved):

PR #309, opened 2021-01-18
chore: clone schemas before normalizing & improve `type` property check

G-Rath:
It makes total sense in hindsight: the replacer gets passed *references* to the properties of the object being stringified, meaning actions on objects and arrays mutate the original schema.

This means currently `ref-common-schema` will end up modifying our schemas even if it doesn't update anything to reference a common schema.

This currently doesn't actually occur as we've already optimized, but making this change now because it's cheap and might as well :)

Also, realised that doing `key === 'type'` is silly because we have schemas that have a property called `type` - while they still will match because we're turning *all* instances of that property into an array, I've adjusted the script to be more correct anyway.

### DIFF
--- a/bin/ref-common-schemas.ts
+++ b/bin/ref-common-schemas.ts
@@ -14,6 +14,9 @@ const metaProperties = [
   "description",
 ];
 
+const cloneObject = <TObject extends object>(object: TObject): TObject =>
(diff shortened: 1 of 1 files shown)


G-Rath:
(actually, I'm going to push up another adjustment to the script, because why not)

wolfy1339 (approved):

PR #316, opened 2021-01-23
feat: author_association common schema

wolfy1339:
Taken from the OpenAPI description of the REST API
Ref: https://github.com/octokit/openapi-types.ts/pull/26

### DIFF
--- a/payload-schemas/schemas/commit_comment/created.schema.json
+++ b/payload-schemas/schemas/commit_comment/created.schema.json
@@ -35,17 +35,7 @@
         "created_at": { "type": "string" },
         "updated_at": { "type": "string" },
         "author_association": {
-          "type": "string",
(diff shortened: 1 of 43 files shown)


oscard0m (approved):

oscard0m:
for future additions we don't have an automatic way to detect this right? @wolfy

wolfy1339:
Not currently. We can't have the script automatically do the changes as the schemas have significant differences, but it could make suggestions for us, and we'd apply them where necessary

oscard0m:
> Not currently. We can't have the script automatically do the changes as the schemas have significant differences, but it could make suggestions for us, and we'd apply them where necessary

Yes. Maybe to open an issue to this repo asking for potential changes. What do you think?

Issue #266, opened 2021-01-13
Use GitHub's OpenAPI spec as source of truth for generating schemas


oscard0m:
The [openapi spec](https://github.com/github/rest-api-description) for the rest api shares a lot of properties with these shemas.
Later on, we could try and use it as the source of truth where possible.

_Originally posted by @wolfy1339 in https://github.com/octokit/webhooks/issues/264#issuecomment-759701042_

gr2m:
Is this still relevant? @wolfy1339

wolfy1339:
It still is relevant, but I don't think we need an issue for it.

We can worry about this whenever GitHub decides to publish schemas for Webhooks themselves along with the OpenAPI spec

PR #317, opened 2021-01-23
feat: augment schemas with information fom the REST API schema

wolfy1339:
Ref: #266


payload-schemas/schemas/common/asset.schema.json
```diff
@@ -18,20 +19,27 @@
...
     "download_count": { "type": "integer" },
-    "created_at": { "type": "string" },
-    "updated_at": { "type": "string" },
+    "created_at": { "type": "string", "format": "date-time" },
+    "updated_at": { "type": "string", "format": "date-time" },
     "uploader": { "$ref": "user.schema.json" }
   },
   "additionalProperties": false,
-  "title": "Asset"
+  "title": "Release Asset"
```
G-Rath, line 44: I think it'd be good to rename the file to `release-asset` too

G-Rath (dismissed):

oscard0m (approved):

oscard0m (approved):

oscard0m (dismissed):

payload-schemas/schemas/common/installation-full.schema.json
```diff
@@ -0,0 +1,85 @@
+{
```
wolfy1339, line 1: Anyone have suggestions for a better file name? I'm not too keen on it as it is right now

oscard0m: * `installation-complete` ?
* `installation-detailed` ?

G-Rath: What's the difference(s) between this and `installation`?

Generally I'd recommend using the difference in the name, i.e `InstallationWith<Property>`.
Depending on what the differences are, it could be better to rename this to `Installation` and rename the existing schema to something else.

wolfy1339: This one is for the `installation` and `installation_repositories` events, the `installation` schema is for every event whenever it's sent by an app

G-Rath: > the installation schema is for every event whenever it's sent by an app

Is that a slimmed down version of the "full" schema? if so, I'd recommend renaming it to `installation-lite` or similar - I think we should be favoring matching the models on the backend for these sorts of schemas first, since we're going to have a number of "lite" versions in events just due to the nature of REST APIs.

wolfy1339: It only has one of the properties from the other `installation` schema, the `id` property, plus an additional property `node_id`

G-Rath: ah yeah - then 💯 this is what I expected.

I'd call this `Installation` and the current `installation` common schema should be something like `InstallationRef`.

(Actually, that's exactly what it should be and what I'm meaning - the underlying model on the backend is `installation`, and elsewhere we have references to that model 🎉. It's what I did for `repo` too)

oscard0m (approved):

G-Rath (approved):

PR #324, opened 2021-01-29
fix: name of `installation` `suspend` & `unsuspend` action schemas

G-Rath:
Follows up #300


wolfy1339 (approved):
Woops :sweat_smile:

This should probably be checked to make sure the `$id` and `title` are the same as the `action`

PR #331, opened 2021-01-29
fix: improve usage of `milestone` common schema

G-Rath:
When present, `tsAdditionalProperties` is used as the value for `additionalProperties` only when compiling a given schema to TypeScript, which allows us to use `allOf` to "extend" a common schema without weakening our types.

This is required because we have to allow additionalProperties in the extending schema type (the schema in index `1` of the `allOf` array) since it's only a subset of the properties defined in the parent schema. However if `additionalProperties` is not set to `false` when generating the types it adds `[key: string]: unknown`, which makes the type an open object which is incorrect.

The drive behind this is that ideally the common schemas/types should represent the *general* state of a particular type rather than the state of the type for only a specific subset of schemas, with the rest having to inline the type as that's the natural thing to expect. (aka, we want common schemas to represent the superset of all possible states of a given common schema)

(this is actually approaching a similar state to what the schemas were originally in with their actions, which isn't a bad thing - it could make sense down the line to make the common schemas even more specific with `oneOf`s, but they're probably not going to be big enough for it to be worth it, vs just using `allOf` inline).

This method of extending should cover most of the common use-cases, so long as the properties being modified exist on the parent schema. I've got ideas on how to improve this which I'll probably land later on :)

### DIFF
--- a/bin/octokit-types.ts
+++ b/bin/octokit-types.ts
@@ -2,8 +2,8 @@
 
 import { strict as assert } from "assert";
 import { promises as fs } from "fs";
-import { JSONSchema7 } from "json-schema";
-import { compileFromFile } from "json-schema-to-typescript";
+import { JSONSchema4, JSONSchema7 } from "json-schema";
+import { compile } from "json-schema-to-typescript";
 import { format } from "prettier";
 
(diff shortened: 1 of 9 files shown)


wolfy1339 (approved):
This is awesome! :heart:

PR #332, opened 2021-01-29
feat: create `Link` interface

G-Rath:
It's a small one, but still nice to have since they should all be the same format etc

### DIFF
--- a/bin/ref-common-schemas.ts
+++ b/bin/ref-common-schemas.ts
@@ -12,6 +12,7 @@ const metaProperties = [
   "$id",
   "title",
   "description",
+  "format",
 ];
 
 const cloneObject = <TObject extends object>(object: TObject): TObject =>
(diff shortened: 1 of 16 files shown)


wolfy1339 (approved):
It's also like this in the REST API OpenAPI schema

G-Rath:
```
❌ Payload 'pull_request/assigned.payload.json' does not match schema
[
  {
    keyword: 'format',
    dataPath: '/pull_request/_links/review_comment/href',
    schemaPath: 'link.schema.json/properties/href/format',
    params: { format: 'uri' },
    message: 'should match format "uri"'
  },
  ```

Interesting...

G-Rath:
@wolfy1339 seems that `href` doesn't have to confirm to uri format, since it can be "https://api.github.com/repos/Codertocat/Hello-World/pulls/comments{/number}" which is not a valid uri.

wolfy1339:
I noticed that I'm many places when adding those in

G-Rath:
It's because it should `uri-template` format instead :)

wolfy1339 (approved):

PR #342, opened 2021-01-30
refactor: move common functions in bin scripts into utils folder

G-Rath:
Have I gone a little overboard? Maybe
Have I got a follow-up PR refactoring argv parsing that's probably overboard? Yup 😁
Does it have `--help` doc support? 💯

----

This does what it says in the title - the majority of the refactoring is moving things into `utils/` wholesale, with a sprinkle of name and code refactoring based on reducing duplication and such.

PR #343, opened 2021-01-31
chore: refactor argv parsing for bin scripts to support `--help` flag

### LINKED ISSUE #314: Add comment description to "extract-common-schema" script
@G-Rath would you mind adding a few lines to the top of [`/bin/extract-common-schema.ts`](https://github.com/octokit/webhooks/blob/master/bin/extract-common-schema.ts) describing what the script does and how it works? No need to go into much detail, just to give a quick overview

G-Rath:
(branched off #342 as it was just easier)

This adds support for `--help` to *all* `bin/` commands. The docs won't be the prettiest right now, and I invite others to help add more info as I've already got a bit of a headache from the initial writing 😅

I'll probably refactor `octokit-webhooks` after this is landed to use `parseArgv` instead of `yargs` for consistency and to simplify things a litte (I've got nothing against `yargs`, but it's a little overkill in this situation I think vs getting the types right 🤷).

Closes #314


gr2m:
> I'll probably refactor `octokit-webhooks` after this is landed to use `parseArgv` instead of `yargs` for consistency and to simplify things a litte (I've got nothing against `yargs`, but it's a little overkill in this situation I think vs getting the types right 🤷).

Good with me. Only reason it currently uses `yargs` is that it's the tool I know. Happy to use `parseArgv` instead. Is that a library? Can you share a link?

G-Rath:
> Is that a library? Can you share a link?

Nope, it's [a utility function](https://github.com/octokit/webhooks/pull/343/files#diff-acbfee01dd846ca93061a5ae27ce28c61ec2e4bd0568861cd4077ec85cc7e16bR106) :)

gr2m (approved):
Really appreciate all the docs!

G-Rath:
No problem! Would you be able to approve #342 first so I can merge that before this one? as I branched off that (which is why the diff is so extreme)

PR #353, opened 2021-02-01
feat: add properties to index.json

wolfy1339:
This is a simple addition that can help in updating schemas automatically with new docs updates

### DIFF
--- a/index.json
+++ b/index.json
@@ -3,6 +3,56 @@
     "name": "check_run",
     "description": "Check run activity has occurred. The type of activity is specified in the `action` property of the payload object. For more information, see the \"[check runs](https://docs.github.com/en/rest/reference/checks#runs)\" REST API.",
     "actions": ["completed", "created", "requested_action", "rerequested"],
+    "properties": {
+      "check_run": {
+        "type": "object",
+        "description": "The [check\\_run](/en/rest/reference/checks#get-a-check-run)."
+      },
+      "check_run[status]": {
+        "type": "string",
+        "description": "The current status of the check run. Can be `queued`, `in_progress`, or `completed`."
+      },
+      "check_run[conclusion]": {
+        "type": "string",
+        "description": "The result of the completed check run. Can be one of `success`, `failure`, `neutral`, `cancelled`, `timed_out`, `action_required` or `stale`. This value will be `null` until the check run has `completed`."
+      },
+      "check_run[name]": {
+        "type": "string",
+        "description": "The name of the check run."
+      },
+      "check_run[check_suite][id]": {
+        "type": "integer",
+        "description": "The id of the check suite that this check run is part of."
+      },
+      "check_run[check_suite][pull_requests]": {
+        "type": "array",
+        "description": "An array of pull requests that match this check suite. A pull request matches a check suite if they have the same `head_sha` and `head_branch`. When the check suite's `head_branch` is in a forked repository it will be `null` and the `pull_requests` array will be empty."
+      },
+      "requested_action": {
+        "type": "object",
+        "description": "The action requested by the user."
+      },
+      "requested_action[identifier]": {
+        "type": "string",
+        "description": "The integrator reference of the action requested by the user."
+      },
+      "repository": {
+        "type": "object",
+        "description": "The [`repository`](/en/rest/reference/repos#get-a-repository) where the event occurred."
+      },
+      "organization": {
+        "type": "object",
+        "description": "Webhook payloads contain the [`organization`](/en/rest/reference/orgs#get-an-organization) object when the webhook is configured for an organization or the event occurs from activity in a repository owned by an organization."
+      },
+      "installation": {
+        "type": "object",
+        "description": "The GitHub App installation. Webhook payloads contain the `installation` property when the event is configured for and sent to a GitHub App."
+      },
+      "sender": {
+        "type": "object",
+        "description": "The user that triggered the event."
+      }
+    },
     "examples": [
       {
         "action": "created",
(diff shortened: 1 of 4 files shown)


oscard0m (commented):
LGTM

gr2m (commented):
Looks great so far 👍🏼

lib/types.ts
```diff
@@ -7,6 +7,7 @@ export interface Webhook<TExample extends object = object> {
   name: string;
   description: string;
   actions: string[];
+  properties: { [key: string]: { type: string; description: string } };
```
G-Rath, line 10:
```suggestion
  properties: Record<string, { type: string; description: string }>;
```

nit: you can use `Record for this

lib/section-to-webhook.ts
```diff
@@ -51,17 +52,40 @@ const getDescription = ($: cheerio.Root): string =>
...
   if (!$table.is("table")) {
     return [];
   }
 
-  const [keyEl, , descriptionEl] = $table
-    .find("tbody tr:first-child td")
+  return $table.find("tbody tr").get() as cheerio.Element[][];
+};
+
+const getProperties = ($: cheerio.Root) => {
```
G-Rath, line 65:
```suggestion
const getProperties = ($: cheerio.Root): Webhook['properties']  => {
```

Would be good to have the return type here so that TS ensures the implementation is returning the correct type (otherwise TS will error on line 38 if the implementation returns something different, which implies the issue is on that line rather than with whats being returned).

(ideally the other methods should have return types too)

lib/section-to-webhook.ts
```diff
@@ -51,17 +52,40 @@ const getDescription = ($: cheerio.Root): string =>
...
+  const propertiesEl = getPropertiesEl($).slice(1);
+  const properties: {
+    [key: string]: { type: string; description: string };
+  } = {};
```
G-Rath, line 69:
```suggestion
  const properties: Webhook['properties'] = {};
```

lib/section-to-webhook.ts
```diff
@@ -51,17 +52,40 @@ const getDescription = ($: cheerio.Root): string =>
...
     return [];
   }
 
-  const [keyEl, , descriptionEl] = $table
-    .find("tbody tr:first-child td")
+  return $table.find("tbody tr").get() as cheerio.Element[][];
+};
+
+const getProperties = ($: cheerio.Root) => {
+  const propertiesEl = getPropertiesEl($).slice(1);
```
G-Rath, line 66:
```suggestion
  const [, ...propertiesEl] = getPropertiesEl($);
```

nit: I don't mind using `slice`, but find using `...` a bit nicer in these situations as you don't have to think about the behaviour of `slice` (i.e "is this taking the first element, or all elements except the first 🤔")

G-Rath (commented):
Looks good! got a few recommendations - feel free to ignore if you're still working on it :)

index.json
```diff
@@ -19887,6 +20468,105 @@
...
+          "46": " ",
+          "47": "`",
+          "48": "e",
+          "49": "d",
+          "50": "i",
+          "51": "t",
+          "52": "e",
+          "53": "d",
+          "54": "`",
+          "55": ".",
```
wolfy1339, line 20534: This is an unintended side effect... How unfortunate that there's a property named `description`.

How should I handle this?

index.json
```diff
@@ -8716,6 +9020,46 @@
     "name": "gollum",
     "description": "A wiki page is created or updated. For more information, see the \"[About wikis](/en/github/building-a-strong-community/about-wikis)\".",
     "actions": [],
+    "properties": {
+      "pages": {
+        "page_name": {
```
wolfy1339, line 9025: I'm not sure how to handle arrays ie `pages[][page_name]`

Any pointers?

index.json
```diff
@@ -49642,6 +50937,114 @@
...
+          "41": " ",
+          "42": "`",
+          "43": "e",
+          "44": "d",
+          "45": "i",
+          "46": "t",
+          "47": "e",
+          "48": "d",
+          "49": "`",
+          "50": ".",
```
wolfy1339, line 50998: When I apply the following diff
```diff
diff --git a/lib/section-to-webhook.ts b/lib/section-to-webhook.ts
index 2a0a264..2fd1a40 100644
--- a/lib/section-to-webhook.ts
+++ b/lib/section-to-webhook.ts
@@ -76,12 +76,13 @@ const getProperties = ($: cheerio.Root): Webhook["properties"] => {
     },
     obj: { [key: string]: any }
   ) => {
-    const matches = [...str.matchAll(/(\w+)/g)];
-    return matches.reduce((cursor, [, key], index) => {
+    const matches = Array.from(str.matchAll(/(\w+)/g), m => m.slice(1,4));
+    return matches.reduce((cursor, [key, array, arrayKey], index) => {
+      const propKey = key ?? arrayKey;
       if (index + 1 === matches.length) {
-        return (cursor[key] = value);
+        return (cursor[propKey] = value);
       }
-      return (cursor[key] = { ...cursor[key] });
+      return (cursor = { 'type': array === '[]' ? 'array' : 'object', ...cursor });
     }, obj);
   };
```
I get the expected result of having a string as the property and not what it currently is `{ ...'string' }`

Any pointers welcome

wolfy1339:
I'm currently having trouble transforming the string notation to JS objects. The current code has some flaws, see my review comments as to what issues i'm getting

wolfy1339:
I reverted this back to my initial commit, and will work on transforming the string notation into object paths later on

index.json
```diff
@@ -3,6 +3,56 @@
     "name": "check_run",
     "description": "Check run activity has occurred. The type of activity is specified in the `action` property of the payload object. For more information, see the \"[check runs](/en/rest/reference/checks#runs)\" REST API.",
     "actions": ["completed", "created", "requested_action", "rerequested"],
+    "properties": {
+      "check_run": {
+        "type": "object",
+        "description": "The [check\\_run](/en/rest/reference/checks#get-a-check-run)."
```
gr2m, line 9: We will need to turn the relative URLs into absolute URLs

gr2m (commented):
other than the relative URLs this looks good to me

index.json
wolfy1339: That is a bug in the script, since they moved domains, the URLs haven't been changed to absolute URLs on our side

wolfy1339: Fix on the way in #363

oscard0m (approved):

PR #356, opened 2021-02-03
fix: mark `author` and `committer` as nullable in `status` schema

G-Rath:


payload-examples/api.github.com/status/1.payload.json
```diff
@@ -0,0 +1,183 @@
+{
```
wolfy1339, line 1: Could you rename this to something more descriptive?
`with-author-committer-null.payload.json`?

wolfy1339 (changes requested):

wolfy1339 (approved):

PR #366, opened 2021-02-08
Update Codescanning Alert Payload Schemas to Include a Sender Field

natashaU:
Hi there 👋 

I'm from the codescanning Team at GitHub and we're doing a slight update. Previously:

sender | object | If the action is reopened_by_user or closed_by_user, the sender object will be the user that triggered the event. The sender object is empty for all other actions.
-- | -- | --

See here for more info: https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#code_scanning_alert

Now, we'll include a default sender object instead of having that field be empty. 

We basically want to update the payload schemas to look like the `closed_by_user` schema by having the sender field be required, so we're just adding two lines of code to all the schemas in the code-scanning-alert folder:

1. https://github.com/octokit/webhooks/blob/03b633b2aed02709c55e3df897747f278ad24397/payload-schemas/schemas/code_scanning_alert/closed_by_user.schema.json#L5

2. https://github.com/octokit/webhooks/blob/03b633b2aed02709c55e3df897747f278ad24397/payload-schemas/schemas/code_scanning_alert/closed_by_user.schema.json#L126

cc/ @bogdanap @anaarmas @krukow

@bogdanap What do you think about defaulting to github as the sender here in the example?

### DIFF
--- a/index.json
+++ b/index.json
@@ -4979,6 +4979,26 @@
         "installation": {
           "id": 1,
           "node_id": "MDIzOkludGVncmF0aW9uSW5zdGFsbGF0aW9uMQ=="
+        },
+        "sender": {
+          "login": "Codertocat",
+          "id": 21031067,
+          "node_id": "MDQ6VXNlcjIxMDMxMDY3",
+          "avatar_url": "https://avatars1.githubusercontent.com/u/21031067?v=4",
+          "gravatar_id": "",
+          "url": "https://api.github.com/users/Codertocat",
+          "html_url": "https://github.com/Codertocat",
+          "followers_url": "https://api.github.com/users/Codertocat/followers",
+          "following_url": "https://api.github.com/users/Codertocat/following{/other_user}",
+          "gists_url": "https://api.github.com/users/Codertocat/gists{/gist_id}",
+          "starred_url": "https://api.github.com/users/Codertocat/starred{/owner}{/repo}",
+          "subscriptions_url": "https://api.github.com/users/Codertocat/subscriptions",
+          "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+          "repos_url": "https://api.github.com/users/Codertocat/repos",
+          "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+          "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+          "type": "User",
+          "site_admin": false
         }
       }
     ]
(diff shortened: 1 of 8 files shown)


wolfy1339:
Do you have an example payload with which we can confirm these changes?

natashaU:
@wolfy1339 

Hey there. I updated the PR description --- will get back to you shortly on the example. Thank you!

wolfy1339:
No problem. I saw the updates.

It's greatly appreciated to have GitHub Staff help out keeping these schemas up to date

wolfy1339:
> If the `action` is `reopened_by_user` or `closed_by_user`, the `sender` object will be the user that triggered the event. The `sender` object is the GitHub organization for all other actions.

@G-Rath @gr2m @oscard0m  Since the sender defaults to the [GitHub org](https://api.github.com/orgs/GitHub) (Only including properties that match the current schemas, the others are additional to the REST API), how do you think we should handle this? Should we keep it using `common/user.schema.json` or be more specific and use the GitHub org values?

G-Rath:
I'm fine with it using the general `User` interface for now, and have it marked up as an extra useful addition to land on its own (so i.e make a new `feature`-tagged issue maybe?)

oscard0m:
> I'm fine with it using the general `User` interface for now, and have it marked up as an extra useful addition to land on its own (so i.e make a new `feature`-tagged issue maybe?)

I'm probably missing context here but, should these changes come from GitHub Docs Open API first?

wolfy1339:
> I'm probably missing context here but, should these changes come from GitHub Docs Open API first?

The changes are coming from the GitHub team responsible for the code scanning event. The changes are getting made to the docs as well.

Where there is currently no `sender` property, there will be one, and it will be the GitHub org values

oscard0m:
> The changes are coming from the GitHub team responsible for the code scanning event. The changes are getting made to the docs as well.
> 
> Where there is currently no `sender` property, there will be one, and it will be the GitHub org values

But then, should we wait for this update on GitHub Docs and it will be scraped?

wolfy1339:
> But then, should we wait for this update on GitHub Docs and it will be scraped?

The schemas will still need to be updated regardless, they aren't automatically updated with docs changes.

oscard0m:
> The schemas will still need to be updated regardless, they aren't automatically updated with docs changes.

True. There's @G-Rath script (manually run) right now?

wolfy1339:
> True. There's @G-Rath script (manually run) right now?

I'm not sure what script you're referencing, but none automatically update the schemas. I don't even think that the payloads get updated from the docs either

natashaU:
Quick Update: We decided to switch out the `sender` in the example reopened alert alert payload to the GitHub Org: https://api.github.com/orgs/GitHub. I'll open this PR and ping you when everything is live! Thanks again :)

gr2m (approved):
Thanks a bunch @natashaU!

PR #369, opened 2021-02-12
feat: use the Github org values for CodeScanningAlert

wolfy1339:
Follow up to https://github.com/octokit/webhooks/pull/366#issuecomment-775609711


payload-schemas/schemas/common/github.schema.json
```diff
@@ -0,0 +1,90 @@
+{
```
G-Rath, line 1: I think a better name for this would be `github-org` (or similar), since `GitHub` by itself is pretty ambogious and we could have similar common interfaces in the future.

payload-schemas/schemas/common/github.schema.json
```diff
@@ -0,0 +1,90 @@
...
+    "public_gists",
+    "followers",
+    "following",
+    "html_url",
+    "created_at",
+    "updated_at",
+    "type"
+  ],
+  "type": "object",
+  "properties": {
```
G-Rath, line 36: I think we shouldn't go overboard with these constants as they could easily change and there's not much upside.
I think the main properties we should have as constants are `login`, `id`, `type`, and the url properties.

G-Rath (changes requested):

payload-schemas/schemas/common/github.schema.json
wolfy1339: I removed some constants from properties that could potentially change

G-Rath: Cheers - theres still a few that I think could change (i.e `twitter_username` and `blog`), but you got the main ones (i.e `public_repos`) so this isn't a blocker for me any more :)

G-Rath (approved):

PR #370, opened 2021-02-12
chore(bin): use npx to call ts-node

wolfy1339:
This makes it so we don't have to rely on people having `node_modules/.bin` in their `PATH`

### DIFF
--- a/bin/diff-interface-schemas.ts
+++ b/bin/diff-interface-schemas.ts
@@ -1,4 +1,4 @@
-#!/usr/bin/env ts-node-transpile-only
+#!/usr/bin/env npx ts-node-transpile-only
 
 import { diffString } from "json-diff";
 import { JSONSchema7 } from "json-schema";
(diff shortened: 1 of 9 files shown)


bin/diff-interface-schemas.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env ts-node-transpile-only
+#!/usr/bin/env npx ts-node-transpile-only
```
G-Rath, line 1: It's not safe to use this syntax, as it's not supported everywhere - it gets interpreted as a single command:

```
/c/Users/G-Rath
❯ cat s.ts
#!/usr/bin/env npx ts-node-transpile-only

console.log('hello');

/c/Users/G-Rath
❯ ./s.ts
/usr/bin/env: ‘npx ts-node-transpile-only’: No such file or directory
```

G-Rath (changes requested):

G-Rath:
[See this stackexchange on why we can't do this](https://unix.stackexchange.com/questions/399690/multiple-arguments-in-shebang).

PR #371, opened 2021-02-13
chore(payloads): add code scanning examples

pholleran:
This PR adds payload examples for the following code scanning webhooks:

* `fixed`
* `closed_by_user`
* `created`


gr2m:
The schema needs to be updated. The `"enterprise"` property is currently not defined for the `code_scanning_alert` event

```
  {
    keyword: 'additionalProperties',
    dataPath: '',
    schemaPath: '#/additionalProperties',
    params: { additionalProperty: 'enterprise' },
    message: 'should NOT have additional properties'
  }
```

wolfy1339:
We only have examples from GitHub, and not GHES, which is why the `"enterprise"` property isn't defined in the schemas.

I think the `"enterprise"` property should be dropped for now and we can add it to all the schemas in a follow up issue

payload-examples/api.github.com/code_scanning_alert/created.payload.json
```diff
@@ -0,0 +1,199 @@
...
+    "gists_url": "https://api.github.com/users/Codertocat/gists{/gist_id}",
+    "starred_url": "https://api.github.com/users/Codertocat/starred{/owner}{/repo}",
+    "subscriptions_url": "https://api.github.com/users/Codertocat/subscriptions",
+    "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+    "repos_url": "https://api.github.com/users/Codertocat/repos",
+    "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+    "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+    "type": "User",
+    "site_admin": false
+  },
```
wolfy1339, line 194: This should be `github` as per the docs
> If the action is `reopened_by_user` or `closed_by_user`, the `sender` object will be the user that triggered the event. The sender object is `github` for all other actions.

payload-examples/api.github.com/code_scanning_alert/fixed.payload.json
```diff
@@ -0,0 +1,218 @@
...
+    "gists_url": "https://api.github.com/users/Codertocat/gists{/gist_id}",
+    "starred_url": "https://api.github.com/users/Codertocat/starred{/owner}{/repo}",
+    "subscriptions_url": "https://api.github.com/users/Codertocat/subscriptions",
+    "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+    "repos_url": "https://api.github.com/users/Codertocat/repos",
+    "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+    "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+    "type": "User",
+    "site_admin": false
+  },
```
wolfy1339, line 213: This should be `github` as per the docs
> If the action is `reopened_by_user` or `closed_by_user`, the `sender` object will be the user that triggered the event. The sender object is `github` for all other actions.

wolfy1339 (dismissed):

payload-examples/api.github.com/code_scanning_alert/fixed.payload.json
pholleran: Thanks for catching this

wolfy1339:
All is good to go (mostly). Waiting on #379 before this can go any further

wolfy1339 (approved):

PR #373, opened 2021-02-16
🚧 chore(schemas): add enterprise schema

wolfy1339:
Taken from payload examples in #371
This is not currently in use, yet. We currently don't have any payloads specific to GHES environments currently, and #371 includes some.


wolfy1339:
This is on-hold for now, until we start adding in versioned types for GHES environments, and decoupling the current types, as discussed in https://github.com/octokit/rest.js/discussions/2050#discussioncomment-505404

PR #374, opened 2021-02-18
feat: add uri/uri-template format in more cases

jablko:
Can we add the uri/uri-template format in the following cases?


gr2m (commented):
Okay with me 👍🏼 We should probably add a check for all properties matching `/(^|_)url$?/` to have the `"format": "uri"` property?

wolfy1339 (approved):

wolfy1339:
> Okay with me 👍🏼 We should probably add a check for all properties matching `/(^|_)url$?/` to have the `"format": "uri"` property? 

That would catch the template URIs as well. Another thing, we'd have to make sure that if the property can be `null` as well (or any other type) that we use an `anyOf`, since the test doesn't like it when a property is null and has a `format` property

PR #375, opened 2021-02-19
refactor: use allOf for forkee property, like #345

jablko:
Can we use
```JSON
      "allOf": [
        { "$ref": "common/repository.schema.json" },
        { "type": "object", "properties": { "fork": { "enum": [true] } } }
      ]
```
for the `forkee` property in `fork/event.schema.json`, like in #345?


bin/utils/normalizeSchema.ts
```diff
@@ -6,6 +6,7 @@ const metaProperties = [
   "$id",
   "title",
   "description",
+  "default",
```
wolfy1339, line 9: Why this change?
If I recall correctly, JSON schemas doesn't have this property

wolfy1339:
Please regenerate the types.
`npm run build:all` will rebuild everything

bin/utils/normalizeSchema.ts
jablko, line 9: "default" is a meta-data annotation? https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.2

payload-schemas/schemas/fork/event.schema.json
```diff
@@ -181,7 +179,8 @@
           },
           "additionalProperties": false
         },
-        "public": { "type": "boolean" }
+        "public": { "type": "boolean" },
+        "organization": { "type": "string" }
```
jablko: With these changes the `forkee` property matches `repository.schema.json` and the `ref-common-schemas.ts` script replaces it with a reference. Only the `"fork": { "enum": [true] }` difference is significant?

bin/utils/normalizeSchema.ts
wolfy1339: We use `draft-07` iirc
I remembered wrong. Thanks though.
I'm just wondering why you added it, since we don't use it anywhere currently

payload-schemas/schemas/fork/event.schema.json
wolfy1339: These changes are fine.

payload-schemas/schemas/fork/event.schema.json
```diff
@@ -5,185 +5,10 @@
...
-            "triage": { "type": "boolean" }
-          },
-          "additionalProperties": false
-        },
-        "public": { "type": "boolean" }
-      },
-      "additionalProperties": false
+      "allOf": [
+        { "$ref": "common/repository.schema.json" },
+        { "type": "object", "properties": { "fork": { "enum": [true] } } }
```
wolfy1339, line 10: You will need to add `"tsAdditionalProperties": false` to the second schema object in the `allOf`

wolfy1339 (changes requested):

bin/utils/normalizeSchema.ts
wolfy1339: Could you revert this change in this PR and we can revisit this in another PR later on

wolfy1339 (approved):
Appart from my comment, this LGTM

payload-schemas/schemas/fork/event.schema.json
jablko: Done. :heavy_check_mark:

It seems annoying that json-schema-to-typescript doesn't automatically exclude members that are outside the intersection, without needing `"additionalProperties": false`?

jablko:
I regenerated the types. :heavy_check_mark: Thanks for your help!

payload-schemas/schemas/fork/event.schema.json
wolfy1339: That's because of how JSON schemas work. I'll refer you to @G-Rath he came up with the solution as to how to "extend" the common schemas in #331
We extended the spec and are using that custom property to tell the compiler to not allow other properties.

bin/utils/normalizeSchema.ts
jablko: I think we do use "default", specifically in `repository.schema.json` https://github.com/octokit/webhooks/blob/933e092af3deff7c734d1a8da43fb7c33c610036/payload-schemas/schemas/common/repository.schema.json#L158

As a result, `ref-common-schemas.ts` doesn't replace `forkee` with a reference without this `metaProperties` change. So I think it's related to this PR and should be included, but give the word and I'll revert it.

payload-schemas/schemas/fork/event.schema.json
jablko: Thanks for the pointer to #331: does https://github.com/jablko/webhooks/blob/7a95e2474b58109bf8fc942d1530e9927e3510e2/schema.d.ts#L1795 make the type an open object? I assume the intersection of an open object and a closed object is a closed object, so `"tsAdditionalProperties": false`/suppressing `[k: string]: unknown` is cosmetic, not semantic?

wolfy1339: Its not just cosmetic. It has semantic value as well.

We use it the same as the `additionalProperties` meta key in JSON schema spec when we don't want additional properties. Its use is only for "extending" schemas.

It tells the test suite and the compiler that it includes the other schema, and that it replaces the properties contained with the more specific values in the second schema, and not to allow additional properties

jablko: #331 says `tsAdditionalProperties` is only used when compiling a schema to TypeScript, and I could be wrong, but I assume e.g. `{ fork: boolean } & { fork: true }` and `{ fork: boolean } & { fork: true, [k: string]: unknown }` are semantically equivalent (both closed objects)?

wolfy1339:
They aren't equivalent. One allows any number of properties, the other only allows known properties

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

jablko:
You're right: the intersection of a closed type and an open type is an open type (in TypeScript). My bad. In that case json-schema-to-typescript should properly translate
```JSON
      "allOf": [
        {
          "properties": {
            "fork": { "type": "boolean" },
            "additionalProperties": false
          }
        },
        { "properties": { "fork": { "enum": [true] } } }
      ]
```
into the intersection of two closed types (vs. one closed and one open)? Because in JSON Schema, "allOf" closed and open is closed.

wolfy1339:
In JSON schema, if you don't specify `"additionalProperties": false`, it will get compiled to a typescript object that allows any number of properties.
```
{
  "type": "object",
  "properties": {
    "fork": { "type": "boolean" },
  },
  "additionalProperties": false
}
```
That would get transpiled to
```
{ fork: boolean; }
```

However, the following examples would give us an object with any number of properties (open object)
```json
{
  "type": "object",
  "properties": {
    "fork": { "type": "boolean" },
  }
}

{
  "type": "object",
  "properties": {
    "fork": { "type": "boolean" },
  },
  "additionalProperties": true
}
```
Would get transpiled to
```ts
{ fork: boolean; [key: string]: unknown }
```

JSON schema defaults to open objects unless you specify "additionalProperties": false

In our case with the allOf,
```json
"allOf": [

  {
    "type": "object",
    "properties": {
      "fork": { "type": "boolean" },
    },
    "additionalProperties": false
  },
  {
    "type": "object",
    
    "properties": {
      "fork": { "type": "boolean", "enum": [true] },
    },
  }
]
```

Would get transpiled to
```ts
{ fork: boolean; } & { fork: true; [key: string]: unknown }
```
And that is not the result we want. To get the result we want, we introduced the "tsAdditionalProperties" meta key to tell it to output a closed object
```json
"allOf": [

  {
    "type": "object",
    "properties": {
      "fork": { "type": "boolean" },
    },
    "additionalProperties": false
  },
  {
    "type": "object",
    
    "properties": {
      "fork": { "type": "boolean", "enum": [true] },
    },
    "tsAdditionalProperties": false
  }
]
```
Would get transpiled to
```
{ fork: boolean; } & { fork: true; }
```

That is the expected behaviour of `json-schema-to-typescript`

payload-schemas/schemas/fork/event.schema.json
G-Rath: They are not: the former says there is only one property (`fork`), the other says there is *at least* `fork` but  could be other properties whose types we don't know.

This has very subtle implications in TypeScript relating to things like passing in computed properties and the compatibility of defined properties (as they all must match their index signature).

It also can degrade IDE autocompletion since they tend to suggest different things based on the fact that it's an open object.

jablko:
Agreed, that's the current json-schema-to-typescript behavior. I think it *should* be `{ fork: boolean; } & { fork: true; }` in both of your "allOf" cases, i.e. if one "allOf" item is closed then all items are closed. That way the TypeScript intersection (closed) matches the JSON Schema "allOf" (closed)?

wolfy1339:
The problem is that you are thinking in Object-oriented way, which the JSON schema spec isn't.
It defaults to open objects.
When you use `allOf`, you are asking it to match each and every schema specified in the array, according to the spec.
Maybe @G-Rath can better explain this to you. But this is the intended behaviour according to the spec

jablko:
```JSON
        {
          "type": "object",
          "properties": { "fork": { "type": "boolean", "enum": [true] } }
        }
```
-> `{ fork: true; [k: string]: unknown; }` is the spec'd behavior, I agree, however
```JSON
      "allOf": [
        {
          "type": "object",
          "properties": { "fork": { "type": "boolean" } },
          "additionalProperties": false
        },
        {
          "type": "object",
          "properties": { "fork": { "type": "boolean", "enum": [true] } }
        }
      ]
```
-> `{ fork: boolean } & { fork: true; [k: string]: unknown; }` is not: the "allOf" is closed, the TypeScript intersection, we (now) agree, is open.

wolfy1339:
As I've said, JSON schema objects are open by default.
An `allOf` does not extend schemas, that is NOT spec behaviour, that is behaviour we introduced ourselves to fit our needs.

Extract from their website:
> It is important to note that the schemas listed in an allOf, anyOf or oneOf array know nothing of one another. While it might be surprising, allOf can not be used to “extend” a schema to add more details to it in the sense of object-oriented inheritance.

I invite you to read this excellent markup of the JSON schema spec on `allOf` [here](https://json-schema.org/understanding-json-schema/reference/combining.html#allof).

***
You are comparing against the schemas seperately, they do not get combined
```json
"allOf": [
  {
    "type": "object",
    "properties": {
      "fork": { "type": "boolean" }
     },
     "additionalProperties": false
  },
  {
    "type": "object",
    "properties": {
      "fork": { "type": "boolean", "enum":[true] }
     },
     "additionalProperties": false
  }
]
```

Which would get compiled to
```ts
{ "fork": boolean } & { "fork": true }
```

and that is the intended behaviour according to the spec.

You want to match `allOf` (`&` operator in TS), the following schemas (TS types)

@G-Rath do you have any better explanations

jablko:
My understanding is that `{ "fork": true, "additional": undefined }` is not an instance of the following schema?
```JSON
      "allOf": [
        {
          "type": "object",
          "properties": { "fork": { "type": "boolean" } },
          "additionalProperties": false
        },
        {
          "type": "object",
          "properties": { "fork": { "type": "boolean", "enum": [true] } }
        }
      ]

wolfy1339:
That is correct. Since it only matches the one schema, and not both

jablko:
And the same value -- `{ "fork": true, "additional": undefined }` -- is assignable to `{ fork: boolean } & { fork: true; [k: string]: unknown; }`

wolfy1339:
Correct.

jablko:
But it's not assignable to `{ fork: boolean } & { fork: true; }`

So `{ "fork": boolean } & { "fork": true }`:
1. might be a more faithful, if less literal, translation?
2. isn't *less* conformant to the JSON Schema spec than the current, `{ fork: boolean } & { fork: true; [k: string]: unknown; }` translation? ... in terms of combined behavior. Piecewise I agree: `{ fork: boolean }` is the correct translation of the first "allOf" term and `{ fork: true; [k: string]: unknown; }` is the correct translation of the second.

wolfy1339:
@G-Rath what do you think of this?

G-Rath:
@wolfy1339 I'm not quite sure what @jablko's point is - it sounds like he's arguing that `json-schema-to-typescript` should be outputting a different type, which is fine except we're not `json-schema-to-typescript` 😅

At the end of the day, the types being generated by `json-schema-to-typescript` are the most accurate representation of what the given JSONSchema validates and accepts for the data. The usability of those types are secondary to that because they're secondary for JSONSchema as well - this is why JSONSchema "allows" you do to `allOf`s with subschemas that have `additionalProperties: false` next to subschemas that define "new" properties,  even though in pretty much every situation it'll result in a schema that never be successfully validated against.

PR #377, opened 2021-02-20
refactor: create AlertInstance schema for CodeScaningAlertEvent#alert#instances

wolfy1339:
…#instances

### DIFF
--- a/payload-schemas/schemas/code_scanning_alert/appeared_in_branch.schema.json
+++ b/payload-schemas/schemas/code_scanning_alert/appeared_in_branch.schema.json
@@ -38,30 +38,7 @@
         },
         "instances": {
           "type": "array",
-          "items": {
-            "type": "object",
-            "required": ["ref", "analysis_key", "environment", "state"],
-            "properties": {
-              "ref": {
-                "type": "string",
-                "description": "The full Git reference, formatted as `refs/heads/<branch name>`."
-              },
-              "analysis_key": {
-                "type": "string",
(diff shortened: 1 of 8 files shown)


G-Rath (approved):
nice!

G-Rath:
er before merging - idk if thats a conventional commit 🤔 

(we should probably check that in CI)

G-Rath:
`refactor` is a valid category - it's the capital letter at the start that I think isn't valid.

wolfy1339:
> er before merging - idk if thats a conventional commit thinking
> 
> (we should probably check that in CI)

I don't think `commitlint` is used here

G-Rath:
> I don't think commitlint is used here

No, but we use semantic-release which... actually that's a good point - it won't error on that sort of thing.

oscard0m:
I think in the past there was [this App](https://github.com/apps/semantic-pull-requests) installed for checking Semantic Commits. On a quick search I did not find anything about it. Did we remove it? @gr2m @wolfy1339

G-Rath:
Personally I've not had a lot of joy with that app, as it means PRs often end up without complete green tickets so the UI makes it look like its not in a good state to be merged.

I pushed up [this branch](https://github.com/octokit/webhooks/tree/add-ci-check-for-conventional-commit) earlier setting up a CI check, which I can PR if people are interested, but we've not had issues so far 🤷 

(The above was me just being forgetful 😅)

oscard0m:
Moved discussion here team 💪🏽 https://github.com/octokit/rest.js/discussions/2018

PR #378, opened 2021-02-21
chore: ensure /(^|_)url$/ properties have uri/uri-template format

jablko:
Add a check that `/(^|_)url$/` properties have the uri/uri-template format https://github.com/octokit/webhooks/pull/374#pullrequestreview-593651478

### DIFF
--- a/bin/optimize-schemas.ts
+++ b/bin/optimize-schemas.ts
@@ -90,6 +90,10 @@ forEachJsonFile(pathToSchemas, (pathToSchema) => {
           ) {
             value.additionalProperties ||= false;
           }
+
+          if (key === "url" || (key.endsWith("_url") && key !== "ssh_url")) {
+            value.format ||= "uri";
+          }
         }
 
         if (value.anyOf) {
--- a/payload-schemas/schemas/deployment_status/created.schema.json
+++ b/payload-schemas/schemas/deployment_status/created.schema.json
@@ -35,7 +35,9 @@
         "creator": { "$ref": "common/user.schema.json" },
         "description": { "type": "string" },
         "environment": { "type": "string" },
-        "target_url": { "type": "string" },
+        "target_url": {
+          "oneOf": [{ "type": "string", "format": "uri" }, { "enum": [""] }]
+        },
         "created_at": { "type": "string" },
         "updated_at": { "type": "string" },
         "deployment_url": { "type": "string", "format": "uri" },
--- a/schema.d.ts
+++ b/schema.d.ts
@@ -1760,7 +1760,7 @@ export interface DeploymentStatusCreatedEvent {
     creator: User;
     description: string;
     environment: string;
-    target_url: string;
+    target_url: string | "";
     created_at: string;
     updated_at: string;
     deployment_url: string;


wolfy1339:
This doesn't really work, I'd rather it just be a check and the tests fail.
We can't just assume it is `uri` rather than `uri-template`

bin/optimize-schemas.ts
```diff
@@ -90,6 +90,10 @@ forEachJsonFile(pathToSchemas, (pathToSchema) => {
           ) {
             value.additionalProperties ||= false;
           }
+
+          if (key === "url" || (key.endsWith("_url") && key !== "ssh_url")) {
```
jablko, line 94: `ssh_url`, e.g. `git@github.com:Octocoders/Hello-World.git`, doesn't match the uri format https://github.com/octokit/webhooks/blob/46983cd8739c706639c066eaed538c768c4d9cc3/payload-examples/api.github.com/repository/created.payload.json#L73

payload-schemas/schemas/deployment_status/created.schema.json
```diff
@@ -35,7 +35,9 @@
         "creator": { "$ref": "common/user.schema.json" },
         "description": { "type": "string" },
         "environment": { "type": "string" },
-        "target_url": { "type": "string" },
+        "target_url": {
+          "oneOf": [{ "type": "string", "format": "uri" }, { "enum": [""] }]
```
jablko, line 39: `deployment_status#target_url` can be "" https://github.com/octokit/webhooks/blob/46983cd8739c706639c066eaed538c768c4d9cc3/payload-examples/api.github.com/deployment_status/payload.json#L30

jablko:
I think this PR will ignore properties that already have the uri-template format?

wolfy1339:
Yes, it does. But the point still stands

G-Rath:
I agree with @wolfy1339 - I think it's worth doing as a "do once" locally, but long-term it represents an incorrect assumption about `uri` vs `uri-format`.

Part of the idea of this script is that it can help take care of some of the boring refactors we'd request in reviews when people add a schema so that contributors don't need to know as much about the subtleties in JSONSchema - so if they submit a new schema without the `format` property, this could set it to `uri` incorrectly.

jablko:
Exceptions to the rule (both uri-template exceptions and those à la `ssh_url`) will fail to validate, so they couldn't be set incorrectly?

wolfy1339:
We don't feel like it's warranted to set a default for possible URI properties, however, adding a test to check that they are set to either `uri` or `uri-template` would be best

G-Rath:
If we have any payloads that fit that - iirc that's how we found out about `uri` vs `uri-template`; originally we had this and then get a new payload example that showed it could be a template.

But that's also adding to the overhead: you're now having to figure out if its you or the tool that's wrong.

> adding a test to check that they are set to either uri or uri-template would be best

That'll put you in a very similar situation, since you'll still need to exclude properties that don't fit the mold, and so then have the overhead each time of having to figure out if those properties are really exceptions or not.

jablko:
I agree this will only work in the presence of a payload example.

wolfy1339:
So we all agree that is is not needed?

wolfy1339:
> Well, my novice impression is that there's good correspondence between schemas and payload examples. Were the schemas originally based on the payload examples?

Yes they originally were based on the examples and the docs.
Originally it was one schema per event and not one schema per action as it is now.

> I've only seen properties that are either always URIs or always URI templates, and in those cases the logic in this PR would work, I think?

As you have pointed out previously, there is some instances that can be empty strings or even SSH URIS

> To @G-Rath's point about overhead, if you add a schema, run optimize-schemas.ts, it adds `"format": "uri"` in a bunch of places, your payload examples fail to validate because e.g. `"https://api.github.com/repos/Octocoders/Hello-World/commits{/sha}" should match format "uri"`, and in those cases you need to manually adjust "uri" -> "uri-template", that seems reasonable? (If I understand your point correctly?) It seems like it would somewhat automate the work of inferring the format from the payload examples, saving some effort and improving consistency.

We want to keep things as simple as possible for new contributors that don't know the intricacies of the JSON schema spec.

> So if the schemas aren't backed up by examples then I agree this PR is a bad idea. Or maybe there's a better approach?

As I've suggested, instead of automatically inferring that it is a URI, we simply log a warning in the tests (but does not make tests fail) that tells users that the property is possibly an URI

wolfy1339:
Closing as per the comments left

PR #379, opened 2021-02-21
fix: adjust GitHubOrg schema

wolfy1339:
These values were added in #369, they match the current docs but don't match the rest of the webhooks.

I believe this was a mistake on their end when they added the example in the docs


wolfy1339:
Waiting for confirmation from staff
https://github.community/t/explanations-for-sender-field-of-codescanningalert/163084?u=wolfy1339

gr2m:
> on their end

who is they? Can you mention them? Maybe they could review your changes?

wolfy1339:
The docs changes were this commit: https://github.com/github/docs/commit/2df6837969ca37b303fbbb3d38db409063d65163

> who is they? Can you mention them? Maybe they could review your changes?

That would be @natashaU

> If the action is reopened_by_user or closed_by_user, the sender object will be the user that triggered the event. The sender object is github for all other actions.

My understanding of the docs right now, is that the sender property is the user that triggered the event.
And for the actions `reopened`, `fixed`, `created`, `appeared_in_branch`, the user that triggered the event is `github`

wolfy1339:
In #366 it was originally changed to the `user` schema, and then later down in the comments, they refer to the [`GET /orgs/github`](https://api.github.com/orgs/github) values, and the [docs](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#code_scanning_alert) had the same values (added in https://github.com/github/docs/commit/2df6837969ca37b303fbbb3d38db409063d65163)

wolfy1339:
I just got confirmation that these are the expected values. https://github.com/github/docs/pull/3855#issuecomment-785201572

gr2m (approved):

PR #383, opened 2021-02-24
fix: cleanup properties in issues and pull_request schemas

wolfy1339:
* Removes the `label` property from other actions then then `labeled` and `unlabeled`
* Removes the `assignees` property from all schemas, it doesn't exist at the top level
* Removes the `assignee` property from other actions then then `assigned` and `unassigned`


gr2m (approved):
great work!

G-Rath (approved):

Issue #362, opened 2021-02-07
Moving `index.json`  to a different package


oscard0m:
**What’s missing?**
Would be nice to not pack examples when publishing this package.

**Why?**
Avoid extra package size 

Context: https://github.com/octokit/webhooks.js/pull/439#issuecomment-774753218

oscard0m:
A scenario coming to my mind which could be worrying is `probot` used in lambda functions (where the size very limited for serverless)

G-Rath:
You beat me to it 😅 I'll paste my issue body here :)

wolfy1339:
Perhaps the package name of `@octokit/webhooks-examples` would be good for this?

G-Rath:
The payloads are *big*, and make up the majority of the published package:

```
webhooks on  master is 📦 v0.0.0-development took 5s
❯ du -hc schema.json schema.d.ts index.json index.d.ts
428K    schema.json
136K    schema.d.ts
3.0M    index.json
0       index.d.ts
3.6M    total
```

They were originally being shipped to power type generation, which is now obsolete as we're shipping JSON schemas and generating more accurate types from those.

Since these types are being used in types that are published in downstream packages, we're wanting to include this package as a production dependenecy meaning all that size will be pulled into the dependency tree when all that's really needed is the far smaller `schema.d.ts`.

Additionally, the payloads are currently being shipped as the "main" entity for this repo, meaning doing `require('@octokit/webhooks-definitions` gives you the payload examples rather than main schema (and likewise for the types), which I think is counter interactive given the package name.

G-Rath:
> A scenario coming to my mind which could be worrying is probot used in lambda functions (where the size very limited for serverless)

For sure - `serverless` in general; i.e [`compare-gh-webhook-to-schema`](https://github.com/G-Rath/compare-gh-webhook-to-schema-function) will have the whole package in production right now (I've not yet configured the exclude to strip it out)

wolfy1339:
For whenever we get around to doing this, we can spin out the payload examples directory into another repo while saving their git history using [this technique](https://medium.com/@ayushya/move-directory-from-one-repository-to-another-preserving-git-history-d210fa049d4b)

gr2m:
For context: https://github.com/octokit/webhooks/ was not meant to be a production dependency, similar to https://github.com/octokit/openapi. Moving the examples out of the repository will make it harder to contribute. I really the fact that anyone can just submit a payload example as a means to reproduce a problem with the current schemas.

[...]

PR #411, opened 2021-03-21
Split components into seperate packages

wolfy1339:
As per our discussion in #362, I started refactoring the repository to move the 3 components of this repo (examples, schemas, types) into seperate packages.

All the build scripts work, all test scripts work except for the `validate` script


package.json
```diff
@@ -6,21 +6,16 @@
...
   "author": "Gregor Martynus (https://github.com/gr2m)",
-  "main": "index.json",
-  "files": [
-    "index.d.ts",
-    "index.json",
-    "schema.d.ts",
-    "schema.json"
-  ],
+  "main": "",
+  "files": [],
```
G-Rath, line 10: For this to be non-breaking, we need to keep these files.

We should be able to easily do this by just having a `cp` step as part of the release flow, that just copies these files out of the new sub-packages, since the actual contents shouldn't have changed.

wolfy1339: That would double the size of the final package though, since it would included in the dependencies and locally in the package.

Should we instead make them 4 seperate packages until a breaking release can be made?

G-Rath: > since it would included in the dependencies and locally in the package.

Then don't include the three new packages as dependencies 😉

wolfy1339:
Everything now works for build & test scripts.

Now to figure out how to get `semantic-release` to publish the 4 packages at the same time

bin/validate-payload-examples.ts
```diff
@@ -3,7 +3,7 @@
 import { DefinedError, ErrorObject } from "ajv";
 import path from "path";
 import { inspect } from "util";
-import { ajv, validate } from "../payload-schemas";
+import { ajv, validate } from "../payload-schemas/index";
```
G-Rath, line 6:
```suggestion
import { ajv, validate } from "../payload-schemas";
```

nit: node imports `index` if it exists when you tell it to import a folder, meaning you can leave this file as is.

G-Rath (approved):
Looks good to me

bin/validate-payload-examples.ts
wolfy1339: Yes, it usually does, but in this case, since there's a `package.json` in the directory with a `main` property, it'll use the `main` property which in this case is `schema.json`.

Which is why I used an explicit import instead of implicit

wolfy1339:
@gr2m When you get the time, this is ready for your review.

wolfy1339:
One thing left, If anyone knows how to integrate `semantic-release` with this setup, please holler. That's all that is missing here

gr2m:
are you still using the `beta` branch for pre-releases? If not, I'd delete the branch and re-create it based on latest `master`, then send this PR against beta, so that we can experiment without publishing to `@latest` on npm.

I'm looking into it, I didn't have a project where I needed multiple releases from a single repository yet

gr2m:
Also it looks like the README has not yet been updated. 

What should the 3 packages be? Can we put them in 3 different directories, and have a `README.md` file for the respective package so that it will show up on npmjs.com?

wolfy1339:
> Also it looks like the README has not yet been updated.

I'll update the README

> Can we put them in 3 different directories

They already are seperated, only the global package is
```
payload-types => @octokit/webhooks-types
payload-schemas => @octokit/webhooks-schema
payload-examples => @octokit/webhooks-examples
./ => @octokit/webhooks-definitions
```

> and have a README.md file for the respective package so that it will show up on npmjs.com?

For sure, that can be done.

wolfy1339:
I have found this `semantic-release` plugin, https://github.com/pmowrer/semantic-release-monorepo
It does look like it could very well help us out

gr2m:
> I have found this `semantic-release` plugin, https://github.com/pmowrer/semantic-release-monorepo
> It does look like it could very well help us out

I didn't use that plugin myself, but it sounds like it supports different versions for the different packages in the repository, which is not something we need in our case. 

I think we could use semantic-release for the root (`@octokit/webhooks-definitions`), use [`semantic-release-plugin-update-version-in-files`](https://github.com/gr2m/semantic-release-plugin-update-version-in-files) to replace the versions in the `payload-*/package.json` files, then run `npm publish` in the `payload-*` folders. That we we'd have a single version for all packages in this repository, with one matching github release and its release notes.

wolfy1339:
The changes are made.

Although there won't be a way to tell npm that this should be a beta, currently

gr2m (approved):
Looks good to me 👍

wolfy1339:
It appears as though, it's not publishing the other packages.

The steps get skipped over

wolfy1339:
There is 2 issues here currently:
1. The packages don't seem to want to be published to npm (possibly due to the fact that we are trying to use the `@beta` tag when `@latest` doesn't exist yet
2. `semantic-release-plugin-update-version-in-files` doesn't seem to be working, there is no mention of it in `semantic-release` logs

gr2m:
I’ll have a look ASAP, tonight or tomorrow

wolfy1339:
I don't think this solution for publishing packages will work as is.
Currently it will publish a release on every single push to either the `master` or `beta` branch. What would be needed is logic similar to the one in `@semantic-release/commit-analyzer` to determine if we need to make a release

PR #420, opened 2021-03-28
fix: remove `description` from objects with `$ref` to stop duplication

G-Rath:
Following on from my comment [here](https://github.com/octokit/webhooks/pull/323#issuecomment-808949949), I realised we can do this instead 🎉 

`json-schema-to-typescript` will pull a `$ref` with a description out into its own interface, duplicating it with a number suffix which makes things confusing.

### DIFF
--- a/bin/octokit-types.ts
+++ b/bin/octokit-types.ts
@@ -56,8 +56,16 @@ const compileSchema = async (): Promise<string> => {
   const schema: JSONSchema4 = JSON.parse(
     await fs.readFile("./schema.json", "utf-8"),
     (key, value: unknown) => {
-      if (isJsonSchemaObject(value) && "tsAdditionalProperties" in value) {
-        value.additionalProperties = value.tsAdditionalProperties;
+      if (isJsonSchemaObject(value)) {
+        // $refs with a description result in a duplicate interface being made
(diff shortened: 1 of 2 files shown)


wolfy1339:
Problem with this solution is that the descriptions get lost when there's multiple uses of the same interface that each have their own description

G-Rath:
The problem with the current situation is that we're creating multiple interfaces that are identical except for the description, giving the implication that they're different when really they're not *and* those interfaces are getting an incorrect description since the descriptions are contextual.

What we want is for the description to be placed on the *property* rather than it's interface, since that's where the context is.
afaik `json-schema-to-typescript` doesn't currently support any way to tell it not to create a new interface in this situation, which I'll follow up with on that repo for a long term solution that'll let us have our cake and eat it too :)

wolfy1339:
> The problem with the current situation is that we're creating multiple interfaces that are identical except for the description

I totally get that.

> What we want is for the description to be placed on the property rather than it's interface, since that's where the context is.

Yes, that is right

> afaik `json-schema-to-typescript` doesn't currently support any way to tell it not to create a new interface in this situation

I guess that will be a tradeoff to make for now.

wolfy1339 (approved):

PR #425, opened 2021-03-30
fix: add missing properties to schemas

wolfy1339:


wolfy1339:
@G-Rath would it be possible to get notifications for when webhooks don't pass schemas?

G-Rath:
@wolfy1339 depends on what sort of notifications you had in mind :)

wolfy1339:
Something where it send a notification containing the webhook payload and the error object from ajv.
While only sending the error once per predetermined length amount of time by testing each error against the one in the "cache".

I realize this might be hard in a server-less environment.

G-Rath:
Yeah that should be doable - it's actually not that hard, especially on AWS (as it's just a dynamodb table) but sadly this is using Azure which itself seems fine but the serverless plugin for it.... less so.

I'll add this to my todo list and try to have a look into it this weekend.

wolfy1339:
> I'll add this to my todo list and try to have a look into it this weekend.
@G-Rath have you had the time to look into it

G-Rath:
@wolfy1339 sadly no, but I'll try to do so right now.

couple of blockers:
  * sadly and somewhat ironically being on azure is making this a lot harder than if it was on aws, so I'm considering if I should move it
  * I've got to consider privacy and IP protection - I'm not super fussed about this, but the function is getting events from the private repos of my work which means the payloads include "stuff that isn't public" (I say it like that because we're not a spy agency or anything, so its not the end of the world if you see project names, but still something I have to do some thinking about to make sure I've not missed anything silly)
  * 
------

having typed that, I'm pretty sure I should be to just report the id of the event that failed which is included in the webhooks history on the GH page 🤔

G-Rath (approved):

G-Rath:
tbh configuration is also the other blocker that needs thinking about - how do you subscribe to notifications about *your* repos failing vs mine.

I'm wondering if building a GH app is the way to go - as I imagine the install flow would let you identify "instances" of the webhook better, and let those installs configure where notifications go to for the repos.

wolfy1339:
> I've got to consider privacy and IP protection - I'm not super fussed about this, but the function is getting events from the private repos of my work which means the payloads include "stuff that isn't public" (I say it like that because we're not a spy agency or anything, so its not the end of the world if you see project names, but still something I have to do some thinking about to make sure I've not missed anything silly)

Most, if not all, payloads in the repository are anonymized using Codertocat/Hello-World values.
If you compared the 2 payloads and the schema, and use the anonymized values instead where they both match the schema it could help alleviate the issues with privacy.
For the other properties that don't match, as long as they have the same type they could be anonymized a bit

We could also make a private repo for this where the payloads are only posted there, which would make sure that nobody could see potentially private information.

> I'm wondering if building a GH app is the way to go - as I imagine the install flow would let you identify "instances" of the webhook better, and let those installs configure where notifications go to for the repos.

Maybe that would be the best choice. It would be simplest

Does anyone else have an opinion or want to chime-in @gr2m @oscard0m

PR #433, opened 2021-04-03
ci(release): `package.json` configuration key is `"release"`, not `"releases"`

gr2m:


wolfy1339 (approved):

wolfy1339:
Can you release the 3 packages to npm manually? That way we can properly test the release workflow

gr2m:
I don't think that is necessary. I think we have to do this:
https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#create-and-check-in-a-project-specific-npmrc-file

```yml
- run "echo //registry.npmjs.org/:_authToken=${secrets.OCTOKITBOT_NPM_TOKEN} > .npmrc"
```

could you try that? I can check back in in 30min

wolfy1339:
That didn't work, I read up a little more on npm, and set the `NPM_CONFIG_TOKEN` env variable for each of the publish commends.

Now that works

gr2m:
okay so what's left is to not run publish when semantic-release did not do a release, right?

We could implement a local plugin that runs on the `"success"` hook? 

I feel really stupid that I can't help better, I've been maintaining semantic-release for years now, but so far successfully avoided a monorepo setup 🤷🏼

wolfy1339:
> okay so what's left is to not run publish when semantic-release did not do a release, right?

Exactly.

> We could implement a local plugin that runs on the "success" hook?

That should work

gr2m:
Do you want to give this a try? I can have a go on Monday myself

PR #436, opened 2021-04-10
feat: split components into seperate packages

### LINKED ISSUE #362: Moving `index.json`  to a different package

wolfy1339:
Fixes #362


wolfy1339:
Made this PR so we could consolidate feedback in one place.

wolfy1339:
> Do you want to give this a try? I can have a go on Monday myself (https://github.com/octokit/webhooks/pull/433#issuecomment-812928108)
I looked into this a bit, and I think I'll leave it up to you @gr2m

package.json
```diff
@@ -59,6 +60,23 @@
   "publishConfig": {
     "access": "public"
   },
+  "release": {
+    "plugins": [
```
gr2m, line 64: Let's be explicit about which branches we release from 

```suggestion
    "branches": ["master", "beta"],
    "plugins": [
```

gr2m (changes requested):
we need to figure out not to publish the packages if there was no release from semantic-release

payload-examples/README.md
```diff
@@ -1,4 +1,266 @@
...
+# Octokit Webhooks
+
+> machine-readable, always up-to-date GitHub Webhooks specifications
+
+![Update status](https://github.com/octokit/webhooks/workflows/Update/badge.svg)
+
+## Download
+
+Download the latest specification at
+[octokit.github.io/webhooks/index.json](https://octokit.github.io/webhooks/index.json)
```
gr2m, line 10: I think this will now be 

```suggestion
[octokit.github.io/webhooks/payload-examples/index.json](https://octokit.github.io/webhooks/payload-examples/index.json)
```

wolfy1339: I also found another place that needed to have the URL updated.

Committed here: https://github.com/octokit/webhooks/pull/436/commits/8c9dee4da0e441fb1cd93c4704eb4c5399597998

wolfy1339:
@gr2m It seems specifying the `npm` plugin for `semantic-release` with different `pkgRoot`s works.

https://github.com/octokit/webhooks/runs/2371208872?check_suite_focus=true

gr2m:
I think the easiest would be to implement an internal semantic-release plugin that acts on the "success" callback, does the `replace` of the `package.json` version in each folder, then runs publish in that folder. I might build this is a separate npm package, seams simple and reusable enough

gr2m:
I think we got it now. In the end, the change was quite straight forward, without the need of a new plugin. In fact, I was even able to remove `emantic-release-plugin-update-version-in-files`, by simply specifying the `npm` plugin to run multiple times, once per package we want to publish. I think this will even result in nicer comments about new releases, because the comment will add one list item per published package.

Using the `npm` plugin takes care of everything we need, including

- set new version in package.json 
- publish to the correct channel based on current branch name

Big shoutout to @pvdlg for this powerful plugin architecture 🎉

gr2m:
> @gr2m It seems specifying the `npm` plugin for `semantic-release` with different `pkgRoot`s works.
> 
> https://github.com/octokit/webhooks/runs/2371208872?check_suite_focus=true

lol are you kidding me, I swear I only saw this now 🤦🏼 I'm sorry I've overwritten your commits now because I'm dumb. 

But hey, we both arrived at the same conclusion, so that's great :)

gr2m (approved):

gr2m:
I guess there is no breaking change because we do not update the existing `@octokit/webhooks-definitions` module? I'll let you do the honors of merging. Once the new packages were released successfully, we should deprecate the existing `@octokit/webhooks-definitions` package with an explanation on how to migrate.

Maybe we could add a section "Upgrade from `@octokit/webhooks-definitions` to the README for some time`? Alternatively we could create an issue and link to that instead

wolfy1339:
@gr2m 
https://github.com/octokit/webhooks/commit/bff89adef9246ded45ad7f894cbbf4de71d11b14 removes the semantic-release configuration for releasing the other packages.

We both missed that 😅

wolfy1339:
@gr2m The release on `master` is failing since the release and tag already exist. Could you delete the existing releases of `v3.68.x`, and manually make a new `v3.68.0` based on `master`?

`semantic-release` failed to create the release because it already exists and future versions exist as well

gr2m:
> @gr2m The release on `master` is failing since the release and tag already exist. Could you delete the existing releases of `v3.68.x`, and manually make a new `v3.68.0` based on `master`?
> 
> `semantic-release` failed to create the release because it already exists and future versions exist as well

I'm on it. What happened is that I mis-configured `branches`. Instead of

```json
    "branches": [
      "master",
      "beta"
    ],
```

we want

```
    "branches": [
      "master",
      {
        "name": "beta",
        "prerelease": true
      }
    ],
```

# 🤦🏼

gr2m:
we should be all good again

PR #437, opened 2021-04-11
fix: `username` property on `Committer` can be not present

wolfy1339:
Caught this while looking through webhooks logs


gr2m:
Can you share how to get into that state? Would it be a commit by a non-github user?

gr2m (approved):

wolfy1339:
> Can you share how to get into that state? Would it be a commit by a non-github user?

Yes, exactly, a commit where the email of the author/committer doesn't match a github user.

wolfy1339:
> Can you share how to get into that state?

In my case, I help run a wiki where the content is synced in a git repo, each edit is one commit. Not everyone has the email associated with their wiki account listed on their github account, therefore there won't be a `username` field.
The pusher (`committer`) doesn't have an email that is listed to a particular github account either, so the `committer` field doesn't have a `username` property

PR #439, opened 2021-04-12
feat: add new & missing properties to events

wolfy1339:
Just a bunch of little patches for new properties, updates to existing values.


gr2m (approved):
Amazing work!

PR #443, opened 2021-04-15
feat: enterprise webhooks examples

### LINKED ISSUE #25: scrape and export GitHub Enterprise versions of webhooks
`https://developer.github.com/enterprise/<GHE version>/v3/activity/events/types/`,
e.g. https://developer.github.com/enterprise/2.16/v3/activity/events/types/

wolfy1339:
Closes #25
Opening this for discussion and feedback.


lib/check-or-update-webhooks.ts
```diff
@@ -10,24 +10,77 @@ import {
...
+    for (let gheVersion of gheVersions) {
+      await checkOrUpdateWebhooks({
+        cached,
+        checkOnly,
+        ghe: gheVersion,
+      });
+    }
+
+    return;
+  }
```
gr2m, line 67: I think I would move the `ghe === ""` check above setting `baseUrl` and `folderName`. I was confused for a second how this works because of the order

gr2m (commented):
This all looks great! Sorry it took me so long to have a look at it. If you ask me, I think we can go ahead and merge it, it won't break anything we already have for api.github.com, but creates an opportunity for others to contribute to the GHES and GHAE webhook examples and schemas

lib/check-or-update-webhooks.ts
wolfy1339: Done.

wolfy1339:
There isn't any schemas in this PR. I'd leave that up for later.

gr2m:
> There isn't any schemas in this PR. I'd leave that up for later.

works for me, let's ship it! Thanks for all the work you put into this! I'm still waiting to hear back about a GHES test instance for you, sorry this takes so long :(

gr2m (approved):

PR #462, opened 2021-04-27
fix: update `d.ts` for `payload-examples` to match the JSON

wolfy1339:


payload-examples/package.json
```diff
@@ -12,7 +12,9 @@
     "index.json"
   ],
   "scripts": {},
-  "dependencies": {},
+  "dependencies": {
+    "@octokit/webhooks-types": "^3.70.0"
```
gr2m, line 16: might this dependency cause us trouble later? Say we release v4, we'll have to remember to bump the version here manually, right?

wolfy1339: We could have `semantic-release` update the version whenever a new release is cut, and make sure that `@octokit/webhooks-types` is published first

gr2m: yeah I think that would be good, if it's not too much trouble to set it up

gr2m (approved):
Looks good 👍🏼 Shall we merge this into `beta` first to see if the release is working out okay?

I'd delete the current beta branch and then re-create it based on latest master, then update the pull request's head to `beta`

wolfy1339:
Yeah, let's test the release process on`beta`

package.json
```diff
@@ -71,22 +72,30 @@
...
       "@semantic-release/release-notes-generator",
       "@semantic-release/github",
+      [
+        "semantic-release-plugin-update-version-in-files",
+        {
+          "files": [
+            "payload-examples/package.json"
+          ]
+        }
+      ],
```
gr2m, line 82: @wolfy1339 can you recall why we added this? I don't think we have to update the version file in `payload-examples/package.json` manually, it will be replaced by the `npm` plugin right before the release?

wolfy1339: It was to update the version of the dependency on `@octokit/webhooks-types`

That way the version would not be out of sync

gr2m: got it thanks!

PR #466, opened 2021-04-27
fix: update `d.ts` for `payload-examples` to match the JSON

wolfy1339:
Release went smoothly with no issues

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ jobs:
         with:
           node-version: 12.x
       - run: npm ci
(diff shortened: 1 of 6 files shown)


wolfy1339:
The release failed on this run because I did a rebase since I had 2 unrelated commits on the branch.
Otherwise it releases fine, and the [`package.json` as seen on unpkg](https://www.unpkg.com/browse/@octokit/webhooks-examples@3.70.1-beta.2/package.json) is clean

gr2m (approved):

PR #468, opened 2021-04-29
fix(types): mark `WebhookDefinition#examples` to be an `Array`

wolfy1339:
The examples property is an Array of `WebhookEventMap[TName]` but it wasn't marked as so


wolfy1339:
This fixes https://github.com/probot/probot/pull/1481#issuecomment-828853658

gr2m (approved):

PR #484, opened 2021-05-16
feat: `PullRequest#body` can be of type `string` or `null`, `Release#{body, name}` are only of type `string`, add new examples for `ReleaseEvent`

wolfy1339:


payload-examples/api.github.com/release/edited.payload.json
```diff
@@ -0,0 +1,164 @@
...
+  "release": {
+    "url": "https://api.github.com/repos/Codertocat/Hello-World/releases/17372790",
+    "assets_url": "https://api.github.com/repos/Codertocat/Hello-World/releases/17372790/assets",
+    "upload_url": "https://uploads.github.com/repos/Codertocat/Hello-World/releases/17372790/assets{?name,label}",
+    "html_url": "https://github.com/Codertocat/Hello-World/releases/tag/0.0.1",
+    "id": 17372790,
+    "node_id": "MDc6UmVsZWFzZTE3MzcyNzkw",
+    "tag_name": "0.0.1",
+    "target_commitish": "master",
+    "name": null,
```
gr2m, line 20: You changed the release name to an empty string, and `release.name` was set to `null`?

gr2m (approved):

payload-examples/api.github.com/release/edited.payload.json
wolfy1339: It seems it's always an empty string now. The `null` was from previous examples already in the repo.
I double checked by creating new releases, and editing them.

gr2m: thanks for double checking!

PR #500, opened 2021-06-09
chore: fix repository JSON contents not being linted

wolfy1339:


.github/workflows/prettier.yml
```diff
@@ -0,0 +1,24 @@
+on:
+  schedule:
+    # https://crontab.guru/#0_0_*_*_7
+    - cron: "0 0 * * 7"
```
wolfy1339, line 4: @gr2m This value doesn't seem to be liked by GitHub Actions.
I used Crontab Guru to create it
```
invalid `cron` attribute "0 0 * * 7"
```

gr2m (dismissed):

gr2m (commented):
why not add the `{payload-examples,payload-schemas}/**/.json` files to the existing `lint` and `lint:fix` scripts?

wolfy1339:
> why not add the `{payload-examples,payload-schemas}/**/.json` files to the existing `lint` and `lint:fix` scripts?

They already are included.

The script makes sure they are consistently formatted.
> The difference between this script and running `prettier --write` on the files directly is that `prettier` preserves multiline objects.

gr2m:
sorry I don't follow. If the files are already formatted when running `npm run lint:fix`, why the extra workflow?

And why the schedule trigger for the extra workflow? Wouldn't it make more sense to run it on push to the default branch?

wolfy1339:
The script formats the JSON files consistently, by stripping them of all whitespace and newlines. So as to let prettier add all the necessary whitespace and newlines where needed
https://github.com/octokit/webhooks/blob/master/bin/docs/format-with-prettier.md

> And why the schedule trigger for the extra workflow? Wouldn't it make more sense to run it on push to the default branch?

I believed it would be too noisy if done on every push, and once a week seemed appropriate. I'm happy to make it run on push

gr2m:
> The script formats the JSON files consistently, by stripping them of all whitespace and newlines. So as to let prettier add all the necessary whitespace and newlines where needed
> https://github.com/octokit/webhooks/blob/master/bin/docs/format-with-prettier.md

got it 👍🏼 

> I believed it would be too noisy if done on every push, and once a week seemed appropriate. I'm happy to make it run on push

maybe instead of a pull request we could just directly push to `main` and use a personal access token from @octokitbot, and make the account an admin on the account?

wolfy1339:
Sure. That sounds alright

gr2m:
I made @octokitbot an admin

gr2m (approved):

PR #504, opened 2021-06-14
ci(prettier): fail git commit silently

gr2m:
This should avoid errors like this:

image

Not sure if there is a more elegant way but this should work

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -20,5 +20,4 @@ jobs:
             git config user.email "33075676+octokitbot@users.noreply.github.com"
             git config user.name "Octokit Bot"
             git add .
-            git commit -m "style: lint JSON files"
-            git push
+            git commit -m "style: lint JSON files" && git push || true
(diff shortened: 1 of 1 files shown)


wolfy1339:
> Not sure if there is a more elegant way but this should work

I did a little research and found this on Stack overflow.
`git diff-index --quiet HEAD` exits with a non-zero exit code when there are changes
Ref: https://stackoverflow.com/questions/3882838/whats-an-easy-way-to-detect-modified-files-in-a-git-workspace

To use that in actions, we would put it in its own step and give it an ID, and use the `-if: failure()` hook.
Ref: https://github.community/t/using-exit-code-of-previous-step/16798

gr2m:
> use the `-if: failure()` hook.

I cannot recall if using the `if: failure()` expression resets the build status, or if it remains a failure, but executes the additional step? Shall we just give it a try?

wolfy1339:
> Shall we just give it a try?

Yeah let's try that and see the results

gr2m:
all set, merge at your convenience

wolfy1339 (approved):

PR #505, opened 2021-06-14
refactor(update-webhooks): adapt for new header layout of docs & update cache

octokitbot:


wolfy1339 (changes requested):
Payload update script seems to need an update. They changed headers type

wolfy1339:
The docs for `GHES-219` and `GHES-220` don't use the updated format.
Because of that the modifications needed to make it work for the other versions break those 2 versions.

gr2m:
> The docs for `GHES-219` and `GHES-220` don't use the updated format.
> Because of that the modifications needed to make it work for the other versions break those 2 versions.

shall we just remove these?

wolfy1339:
> shall we just remove these?

Sounds reasonable

wolfy1339 (approved):

PR #506, opened 2021-06-15
feat: remove GHES 2.19 and GHES 2.20

wolfy1339:
The docs for `GHES-219` and `GHES-220` don't use the updated format.
Because of that the modifications needed to make it work for the other versions break those 2 versions.

_Originally posted by @wolfy1339 in https://github.com/octokit/webhooks/issues/505#issuecomment-861155865_

Because of the above problem, removing these versions of ghes was seen as the best soution


gr2m (approved):

gr2m:
I'd suggest we remove it as a breaking release?

PR #513, opened 2021-06-21
🤖📯 Octokit webhooks changed

octokitbot:


wolfy1339 (changes requested):
Something has gone wrong...
The payload examples for `api.github.com` have all been removed from the respective `index.json`

wolfy1339:
It seems GitHub has done some more refactoring of it's docs.
They have moved the TOC outside of `#article-contents`.

So now we need the parent of the parent of `#article-contents` to get the same HTML cache.

***

There are also some new properties that have been added to the docs, that need to be added to the in-repo examples & schemas

PR #514, opened 2021-06-21
fix(api.github.com): add new properties for the Container registry

wolfy1339:
GitHub refactored the docs and moved the HTML we want out of that ID. It is another layer down compared to before

closes #513


gr2m (approved):

gr2m:
Should there be changes in the cached HTML as part of this PR?

gr2m:
I'm running the build scripts locally and will push the changes hold on

gr2m:
Sorry now I get that you planned to merge the build changes separately and then update https://github.com/octokit/webhooks/pull/513. I've commited the changes now here, so I think this PR supersedes https://github.com/octokit/webhooks/pull/513

wolfy1339:
There were some other changes I wanted to push regarding updating schemas (if needed) and updating other in-repo examples.

PR #529, opened 2021-07-22
fix(api.github.com): update descriptions for `issues.labeled` and `issues.unlabeled` events

gr2m:
If I understand it correctly, the `issues.labeled` event only adds a label. If a label is removed, the action is set `unlabeled`. Or am I missing something?


wolfy1339:
Yup.
`labeled` is when you add a label to an issue
`unlabeled` is when you remove a label from an issie

gr2m:
I updated descriptions in both the `issues.labeled` and `issues.unlabeled` event

wolfy1339 (approved):

PR #534, opened 2021-08-05
feat: add `workflow_job` schema

G-Rath:
I've taken some liberties with `steps` in terms of being strict on the number of items - I think it work and be fine, but wouldn't be surprised if we later find a payload with more items for an `in_progress` 🤷


gr2m (approved):
👍🏼

wolfy1339:
Looks like you forgot to generate `index.json`

gr2m:
> Looks like you forgot to generate `index.json`

can we enforce it somehow as part of the CI?

wolfy1339:
It can be done easily. There's already a `--check` flag implemented.

G-Rath:
Yeah I thought we already did check it in CI - maybe it got lost in the shuffle when we split the monopackage up?

wolfy1339:
I don't think there was ever a check for the examples

npalm:
Looks like that the job status `queued` is missing

gr2m:
can you file a dedicated issue if the schemas/types are not complete?

PR #535, opened 2021-08-06
ci: test if payload examples for `api.github.com` need to be regenerated

wolfy1339:
As said in https://github.com/octokit/webhooks/pull/534#issuecomment-894025397 add a check to the CI for the example payloads


wolfy1339:
The script doesn't seem to add events that aren't present in the docs

wolfy1339:
Regarding the events that aren't present in github docs: I filed https://github.com/github/docs/issues/9964

wolfy1339:
This is good to merge.

gr2m (approved):

PR #576, opened 2021-10-11
Schema updates

wolfy1339:
- feat: new `keys` permission for apps
- feat: new `run_attempt`, `run_started_at` and `previous_attempt_url` properties for `workflow_run`
- feat: new `topics` property for `repository`
- feat: new `run_attempt` property on `workflow_job`
- feat: new `allow_update_branch` property on `Repository`
- feat: new `draft` property on `issue` object
- feat: `visibility` is a required field on `repository`
- fix: repository is not always present for `project_card` events
- fix: add `internal` to allowed values for `repository#visibility`
- fix: the property `check_suite#after` can be `null`
- fix: add missing `branch_protection_rule` event to `App`
- fix: correct typo in `enum` for `App` key `permission`


G-Rath (approved):
This looks fine to me, but hard to review in detail because of the size - if you can be bothered it could be good to break out into smaller PRs? e.g. `topics` on repositories can definitely be landed I think.

wolfy1339:
> This looks fine to me, but hard to review in detail because of the size - if you can be bothered it could be good to break out into smaller PRs? e.g. `topics` on repositories can definitely be landed I think.

Even splitting it wouldn't really help with the size of the PR that much, as there's updates to almost every payload due to new required properties in the `repository` schema.

PR #588, opened 2021-11-18
build: fix getting sections out of the webhooks docs

octokitbot:


wolfy1339:
GitHub seems to have changed the format of the docs again.

The only thing that seemed to have broken is getting the sections of the Webhooks docs, that code lies in [`lib/get-sections.ts`](https://github.com/octokit/webhooks/blob/master/lib/get-sections.ts).

Instead of a tree structure resembling `div#in-this-article > ul > a`, it is now `div > div#in-this-article, ul > a`.
Each item is now contained in it's own `<ul>` element instead of an `<li>` element (which breaks the HTML semantics), and there is no containing `<ul>` element anymore.
The element with the `id` of `in-this-article` has been moved to the same level as the list of sections in a new "container", thus we have to get it's parent element, and then find every `<a>` element (links) contained within a `<ul>` element.

Relevant code that broke:
https://github.com/octokit/webhooks/blob/master/lib/get-sections.ts#L8-L10

Change needed:
```diff
diff --git a/lib/get-sections.ts b/lib/get-sections.ts
index 56a7f0d..b408ea7 100644
--- a/lib/get-sections.ts
+++ b/lib/get-sections.ts
@@ -6,8 +6,8 @@ export const getSections = (html: string): Section[] => {
   const $ = cheerio.load(html);
 
   const sections = $("#in-this-article")
-    .next("ul")
-    .find("a")
+    .parent()
+    .find("ul a")
     .map((index, element) => {
       const url = $(element).attr("href");
```

wolfy1339 (approved):

PR #597, opened 2021-12-10
refactor: use `got` with retries instead of `@octokit/request` to retrieve the docs HTML pages

gr2m:
will hopefull work around build errors such as
https://github.com/octokit/webhooks/runs/4486395928\?check_suite_focus\=true\#step:7:39

I see these build errors occur several time each day, it's creating quite some noise in my notifications

### DIFF
--- a/lib/get-html.ts
+++ b/lib/get-html.ts
@@ -1,4 +1,4 @@
-import { request } from "@octokit/request";
+import got from "got";
 import cheerio from "cheerio";
 import prettier from "prettier";
 import { State, cache } from ".";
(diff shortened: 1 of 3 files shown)


G-Rath:
~fwiw Sindresorhus writes his packages in ESM modules, meaning this will lock us in to got v11 until we're ready to convert over to ESM ourselves - I don't think it'll be a big blocker, but if...~

Actually, we're using this in one place which is async anyway so we can just do `await import('got')`.

G-Rath (approved):

PR #607, opened 2022-02-01
fix: type override masks the props

y-nk:
The 2nd type definition is masking the 1st one for some reason. The union doesn't seem to work and the props are rather "lost".

image

The only way to get access to `requested_reviewer` is to remove the union and reunite the types.

### DIFF
--- a/payload-types/schema.d.ts
+++ b/payload-types/schema.d.ts
@@ -270,25 +270,13 @@ export type PullRequestReviewRequestRemovedEvent =
        */
       number: number;
       pull_request: PullRequest;
-      requested_reviewer: User;
+      requested_reviewer?: User;
+      requested_team?: Team;
       repository: Repository;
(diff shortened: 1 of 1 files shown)


G-Rath:
`schema.d.ts` is automatically generated from the JSON schemas, so this needs to be fixed there - the changes you've made will otherwise be undone when they're next changed.

You are right though - these types should not be a union in their current state, so this is a bug (hopefully one we can fix, but it might be tricky due to how the type generation works).

In the meantime as a workaround, you can make TypeScript happy by checking for those properties before accessing them: 

```
import { PullRequestReviewRequestedEvent } from '@octokit/webhooks-types'

const fn = (payload: PullRequestReviewRequestedEvent) => {
  if('requested_team' in payload) {
    payload.requested_team
  }
}
```

PR #610, opened 2022-03-07
build: update webhooks docs URL

wolfy1339:


wolfy1339:
The update workflow is currently broken without this patch.

G-Rath (approved):

PR #641, opened 2022-06-15
build: cache updates

octokitbot:
I found new changes on [GitHub’s Webhooks Event Types & Payloads](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads) and want to let you know about it 👋🤖

I can't tell if the changes are fixes, features or breaking changes. You'll have to figure that out on yourself and adapt the commit messages accordingly to trigger the right release, see [our commit message conventions](https://github.com/octokit/webhooks/blob/master/CONTRIBUTING.md#merging-the-pull-request--releasing-a-new-version).

PR #645, opened 2022-06-16
docs(CONTRIBUTING): add section on how to create schemas

### LINKED ISSUE #643: Document how we get from a new webhook event to a new schema and to new types
Follow up to https://github.com/octokit/webhooks/pull/641#discussion_r898325831

wolfy1339:
These are basic steps to creating a schema.
Fixes #643

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,8 +43,8 @@ tests will fail.
 
 The webhook schemas are [JSON schemas](https://json-schema.org/) that are manually created using example payloads for webhook events.
 
-- Create a new folder for the event, if it doesn't already exist
-- Create a new JSON file named `<action>.schema.json` if the event has an `action` property, otherwise name it `event.schema.json`
+- Create a new folder for the event in `payload-schemas/api.github.com/`, if it doesn't already exist
+- Create a new JSON file in that folder named `<action>.schema.json` if the event has an `action` property, otherwise name it `event.schema.json`
 - The `$id` property should start with the event name, followed by the `action` name separated with `$` if the event has an `action` property, e.g. `pull_request$created`. If the event doesn't have an `action` property then simply use `event` instead of the `action` name.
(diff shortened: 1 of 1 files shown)


wolfy1339:
Blocked by #641, as the current cache has multiple entries for the new `security_and_analysis` event, and the `index.json` file didn't get generated in #634 which causes the tests to fail.

Would need to override the required tests for merge.

CONTRIBUTING.md
```diff
@@ -39,6 +39,52 @@ When you send a pull request, make sure that the `index.json` file is up-to-date
 with the changes in the `payload-examples/api.github.com/` folder, otherwise the
 tests will fail.
 
+## Creating a new schema
+
+The webhook schemas are [JSON schemas](https://json-schema.org/) that are manually created using example payloads for webhook events.
+
+- Create a new folder for the event, if it doesn't already exist
+- Create a new JSON file named `<action>.schema.json` if the event has an `action` property, otherwise name it `event.schema.json`
+- The `$id` property's value should be in the format `<eventName>$action`, where `<eventName>` is the name of the event, and `action` is the value of the `action` property of the webhook payload or simply `event` if there is no `action` property for the event
```
G-Rath, line 48: For consistency I think we should use `<action>` here:

```suggestion
- The `$id` property's value should be in the format `<eventName>$<action>`, where `<eventName>` is the name of the event, and `<action>` is the value of the `action` property of the webhook payload or simply `event` if there is no `action` property for the event
```

nit: though I'd probably suggest rephrasing this to be something like:

> The `$id` property should start with the event name, followed by the `action` name separated with `$` if the event has an `action` property, e.g. `pull_request$created`

G-Rath (approved):

CONTRIBUTING.md
wolfy1339: I went with your rephrased version

wolfy1339:
I believe this is ready to merge. (The test failures are unrelated to this PR)

gr2m (approved):
This is really great, thank you  so much for putting it together!

oscard0m:
Arrived late to the review but thanks for this @wolfy1339 ! ❤️ 👏🏽

PR #646, opened 2022-06-16
refactor(update-webhooks): remove classes from elements

### LINKED ISSUE #642: Automated update PRs: reduce noise of cache-only updates
The HTML of docs seems to update much more often then it used to. It would be a good time invest to reduce the noise resulting from that. We could normalize the HTML code before caching it, or avoid cache-only update commits, see
https://github.com/octokit/webhooks/pull/641

wolfy1339:
This change will remove all the classes from the cached HTML, except for classes which we actually look for in our code in order to extract payload examples

Fixes #642


wolfy1339:
I'd like to wait on #641 before merging this, in order to create the least friction/conflicts between PRs

lib/get-html.ts
```diff
@@ -29,7 +29,14 @@ export const getHtml = async (
 
   // get only the HTML we care about to avoid unnecessary cache updates
   $('[data-testid="callout"]').remove();
-  const html = $("#article-contents").parent().parent().html() ?? "";
+  const data = $("#article-contents").parent().parent();
+  data.find("*").each((i, el) => {
+    const classes = $(el).attr("class");
+    if (classes && !classes.includes("language-json") && !classes.includes("warning")) {
```
G-Rath, line 35: Might as well default to an empty array to save a check (|| save using optional chain instead of the check)

```suggestion
    const classes = $(el).attr("class") ?? [];
    if (!classes.includes("language-json") && !classes.includes("warning")) {
```

G-Rath (approved):
awesome!

lib/get-html.ts
G-Rath: huh weird that it makes typescript unhappy - @wolfy1339 feel free to just go with what you originally had 🤷

oscard0m (approved):

lib/get-html.ts
```diff
@@ -29,7 +29,14 @@ export const getHtml = async (
...
   $('[data-testid="callout"]').remove();
-  const html = $("#article-contents").parent().parent().html() ?? "";
+  const data = $("#article-contents").parent().parent();
+  data.find("*").each((i, el) => {
+    const classes = $(el).attr("class") ?? "";
+    if (!classes.includes("language-json") && !classes.includes("warning")) {
+      $(el).removeAttr("class");
+    }
+  });
+  const html = data.html() ?? "";
```
oscard0m, line 39: I think we are missing the use-case where an element has the className `language-json hsjfh` or `warning aksjdhksadj`, in these cases we would be having the same noise since we are not pruning thee hash for those HTMLElements.

I drafted something like this:

```suggestion
  const data = $("#article-contents").parent().parent();
  data.find("*").each((i, el) => {
    const classes = $(el).attr("class");

    if (classes) {
      console.log(classes);
      const filteredClasses = classes
        .split(" ")
        .filter((classSelector: string) =>
          ["language-json", "warning"].includes(classSelector)
        );

      if (filteredClasses.length) {
        $(el).attr("class", filteredClasses.join(" "));
      } else {
        $(el).removeAttr("class");
      }
    }
  });
  const html = data.html() ?? "";
```

oscard0m (changes requested):

lib/get-html.ts
wolfy1339: I just double checked the cache, and the `language-json` class is not present anymore. It seems that part of the code is out of date

oscard0m: > I just double checked the cache, and the `language-json` class is not present anymore. It seems that part of the code is out of date

I would move this to a different PR. They are unrelated changes and in case we want to revert one thing does not imply we want to revert the other.

wolfy1339: All done

oscard0m (approved):

lib/get-html.ts
```diff
@@ -29,7 +29,25 @@ export const getHtml = async (
...
+          ["language-json", "warning"].includes(classSelector)
+        );
+
+      if (filteredClasses.length) {
+        $(el).attr("class", filteredClasses.join(" "));
+      } else {
+        $(el).removeAttr("class");
+      }
+    }
+  });
```
gr2m, line 49: could you add a comment about why we are doing this, maybe with a reference to https://github.com/octokit/webhooks/issues/642?

wolfy1339: Added a comment 510feacc3be13e095d372a93e219827644a0d42a

PR #652, opened 2022-06-28
fix: update selector for payload examples

wolfy1339:
I was looking at the cache when I was working on #646, and noticed there wasn't any usage of the class `language-json` anymore.

Looking at the docs' page source code in the browser, you can notice that `language-json` is only added when JavaScript is enabled.


wolfy1339:
Will require the removal of old GHES versions

wolfy1339:
This is all ready and dandy!

gr2m (approved):
# :shipit:

PR #653, opened 2022-06-28
feat(GHES): add GHES 3.4 and 3.5, remove GHES 2.21, 2.22, 3.0

wolfy1339:
BREAKING CHANGE: Drop GHES 2.21, 2.22, 3.0


nickfloyd (approved):
Looks good!  Just a side note v3.1 was [discontinued](https://docs.github.com/en/enterprise-server@3.1/rest/branches/branches) on 2022-06-03

PR #657, opened 2022-07-05
fix: refactor `workflow_job.queued` event to use the `WorkflowJob` common shema

Helcaraxan:
My personal code generation tool led me to notice a missing refactor for the `queued` variant of the `workflow_job` event. Am refactoring it here along the same lines as what is used for other variants of the same event.

This does require the relaxation of the `minItems` constraints on the `steps` field in the refactored Workflow Job spec and hence results in a slight alteration of the validation semantics of the other event variants, without it being a breaking change.

If someone could point out a way to forgo this constraint relaxation I am all ears but I did try to add an override for `steps` alongside the `status` one within the `queued` spec but this resulted in test failures.

Example, adding the following in `queued.schema.json`:
```json
        {
          "type": "object",
          "required": ["status"],
          "properties": {
            "status": { "type": "string", "enum": ["queued"] },
            "steps": {
              "type": "array",
              "items": { "$ref": "common/workflow-step.schema.json" }
            }
          },
          "tsAdditionalProperties": false
        }
```
results on `npm run test` in:
```text
✅ Payload 'workflow_job/completed.failure.with-organization.payload.json' matches schema
✅ Payload 'workflow_job/completed.success.with-organization.payload.json' matches schema
✅ Payload 'workflow_job/in_progress.payload.json' matches schema
❌ Payload 'workflow_job/queued.payload.json' does not match schema
[
  {
    instancePath: '/action',
    schemaPath: '#/properties/action/enum',
    keyword: 'enum',
    params: { allowedValues: [ 'completed', 'in_progress' ] },
    message: 'must be equal to one of the allowed values'
  },
  {
    instancePath: '/workflow_job/steps',
    schemaPath: '#/properties/steps/minItems',
    keyword: 'minItems',
    params: { limit: 1 },
    message: 'must NOT have fewer than 1 items'
  },
  {
    instancePath: '',
    schemaPath: '#/oneOf',
    keyword: 'oneOf',
    params: { passingSchemas: null },
    message: 'must match exactly one schema in oneOf'
  }
]
```

### DIFF
--- a/payload-schemas/api.github.com/common/workflow-job.schema.json
+++ b/payload-schemas/api.github.com/common/workflow-job.schema.json
@@ -43,8 +43,7 @@
     },
     "steps": {
       "type": "array",
-      "items": { "$ref": "workflow-step.schema.json" },
-      "minItems": 1
+      "items": { "$ref": "workflow-step.schema.json" }
     },
     "conclusion": {
       "type": ["string", "null"],
--- a/payload-schemas/api.github.com/workflow_job/queued.schema.json
+++ b/payload-schemas/api.github.com/workflow_job/queued.schema.json
@@ -10,55 +10,17 @@
     "repository": { "$ref": "common/repository.schema.json" },
     "sender": { "$ref": "common/user.schema.json" },
     "workflow_job": {
-      "type": "object",
-      "required": [
-        "id",
-        "run_id",
-        "run_url",
-        "run_attempt",
-        "node_id",
-        "head_sha",
-        "url",
-        "html_url",
-        "status",
-        "conclusion",
-        "started_at",
-        "completed_at",
-        "name",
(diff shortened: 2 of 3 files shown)


wolfy1339:
This would require some fancier work to get the validation to work.

https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else

You can't really extend schemas in the way you would expect in object-oriented programming. Each schema defined in an `allOf` must pass independently of one another, which is why the `minItems` needs to be relaxed

Helcaraxan:
@wolfy1339 thank you very much for the insight. I understand it better now.

Despite the relaxation is this PR something that you would entertain as a refactor or would it be better to close and keep things as they are right now?

wolfy1339 (approved):

Issue #689, opened 2022-08-09
Changes object has 19 possible keys but type is defined with only 3 of them


LightningStairs:
The `changes` object type is defined to have 3 possible keys: `authorized_actors_only`, `authorized_actor_names`, and `required_status_checks`. But it can have several others as well: 

```
     pull_request_reviews_enforcement_level
      required_approving_review_count
      dismiss_stale_reviews_on_push
      require_code_owner_review
      authorized_dismissal_actors_only
      ignore_approvals_from_contributors
      required_status_checks_enforcement_level
      strict_required_status_checks_policy
      signature_requirement_enforcement_level
      linear_history_requirement_enforcement_level
      admin_enforced
      allow_force_pushes_enforcement_level
      allow_deletions_enforcement_level
      merge_queue_enforcement_level
      required_deployments_enforcement_level
      required_conversation_resolution_level
```

Because of how this type is defined it is only possible to test cases when `authorized_actors_only`, `authorized_actor_names`, and `required_status_checks` change.

https://github.com/octokit/webhooks/blob/e4bef38ae171f929eb4da9d2913f164eedf58a9a/payload-types/schema.d.ts#L793-L803

wolfy1339:
Can you confirm wether this is for GitHub.com and not GitHub.AE or GitHub Enterprise?
Please provide an example payload with those keys

JamieMagee:
The `required_conversation_resolution_level` now has a different type in the `BranchProtectionRule`

https://github.com/octokit/webhooks/blob/5d5bb3ca635b7224a591898321d1a5bad67af3b2/payload-types/schema.d.ts#L440

and `BranchProtectionRuleEditedEvent`

https://github.com/octokit/webhooks/blob/5d5bb3ca635b7224a591898321d1a5bad67af3b2/payload-types/schema.d.ts#L824-L826

wolfy1339:
I think it might be easiest to introduce a new common schema for each rule type.

* `BranchProtectionRuleBinary` - rules where only boolean values are used
* `BranchProtectionRuleMultiLevel` - rules where you can select one of `off`, `non_admins`, or `everyone`
* `BranchProtectionRuleArray` - rules where an array of strings is expected, used for actor and build rules

JamieMagee:
That sounds reasonable, and it's effectively what I did in the .NET version. I think you will also need a `BranchProtectionRuleNumber` as well though:

[...]

PR #690, opened 2022-08-09
fix: `changes` property can be undefined in `branch_protection_rule#edited`, add missing keys to `changes` property in `branch_protection_rule#edited`, new `blocking` permission for apps

### LINKED ISSUE #688: Changes object can be undefined
We we imported the BranchProtectionRuleEditedEvent recently and we came across an issue. The changes object can be undefined but the Type definition doesn't allow for it to be undefined. To replicate this, go to a repo's branch protection settings and it save without changing any of the settings. This fires off a BranchProtectionRuleEditedEvent without the changes object. 

https://github.com/octokit/webhooks/blob/e4bef38ae171f929eb4da9d2913f164eedf58a9a/payload-types/schema.d.ts#L793-L803

wolfy1339:
Fixes #688
Fixes part of #689

### DIFF
--- a/payload-examples/api.github.com/branch_protection_rule/edited.payload.json
+++ b/payload-examples/api.github.com/branch_protection_rule/edited.payload.json
@@ -28,7 +28,20 @@
   },
   "changes": {
     "authorized_actors_only": { "from": false },
-    "authorized_actor_names": { "from": [] }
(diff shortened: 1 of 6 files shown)


gr2m (commented):
wow, lots of new properties, thanks!

gr2m (approved):

PR #693, opened 2022-08-11
fix: correct `enum` values for `required_conversation_resolution_level`, add enum values for `PullRequestReview#state`

wolfy1339:


payload-schemas/api.github.com/pull_request_review/submitted.schema.json
```diff
@@ -37,7 +37,10 @@
           "description": "A commit SHA for the review."
         },
         "submitted_at": { "type": "string", "format": "date-time" },
-        "state": { "type": "string" },
+        "state": {
+          "type": "string",
+          "enum": ["commented", "changes_requested", "approved"]
```
nickfloyd, line 42: I could never find a definitive list for this enum from the [docs](https://docs.github.com/en/rest/pulls/reviews). After parsing it a bit I found examples using both `changes_requested` and `approved`, but I could never find `comment` - would you happen to be able to point me to where you found this list (schema or docs). 

Definitely not a show stopper or anything - I'd just like to find out why I am missing this detail.

nickfloyd (approved):
Just a quick ask about pull request review values... I just couldn't locate the list.

payload-schemas/api.github.com/pull_request_review/submitted.schema.json
wolfy1339: I found it just by comparing payloads of webhooks I received.

A lot of things in these schemas aren't really described in the docs, and we have to infer from the payloads we receive

nickfloyd: Got it... thanks for the clarity.

PR #696, opened 2022-08-11
fix: refactor `BranchProtectionRule` to have a common schema for the different rule types

wolfy1339:
https://github.com/octokit/webhooks/issues/689#issuecomment-1211422677

### DIFF
--- a/payload-schemas/api.github.com/branch_protection_rule/edited.schema.json
+++ b/payload-schemas/api.github.com/branch_protection_rule/edited.schema.json
@@ -14,7 +14,11 @@
         "admin_enforced": {
           "type": "object",
           "required": ["from"],
-          "properties": { "from": { "type": "boolean" } },
+          "properties": {
+            "from": {
+              "$ref": "common/branch-protection-rule-boolean.schema.json"
+            }
+          },
           "additionalProperties": false
         },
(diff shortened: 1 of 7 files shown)


payload-schemas/api.github.com/branch_protection_rule/edited.schema.json
```diff
@@ -14,16 +14,24 @@
         "admin_enforced": {
           "type": "object",
           "required": ["from"],
-          "properties": { "from": { "type": "boolean" } },
+          "properties": {
+            "from": {
+              "$ref": "common/branch-protection-rule-boolean.schema.json"
+            }
+          },
```
oscard0m, line 21: I see the benefit of moving the `enum` in `branch-protection-rule-enformcement-level.schema.json` to centralize the possible values in one place and maintainability.

What is the benefit of doing the same for `integer`, `boolean` and `string[]` here? For sure there is a reason but I'm not understanding it.

wolfy1339: There isn't really that much of a benefit. It's is inline with `octokit/webhooks.net` though

oscard0m (approved):

PR #751, opened 2022-12-15
🤖📯 Octokit webhooks changed

octokitbot:

If there are any new events added, follow the [steps outlined](https://github.com/octokit/webhooks/blob/master/CONTRIBUTING.md#creating-a-new-schema) in order to create the required schemas.
If there were only description updates, or new properties added, simply modify the appropriate schemas and commit them to this PR. Don't forget to generate the types as well.


wolfy1339 (changes requested):
The docs format has changed, the script needs to be updated to try to restore the most functionality

wolfy1339:
With the changes to the docs, there are no longer any example payloads available.

I think the automatic updater should be retired, and a new script that opens an issue when a new event is found be put in it's place instead.

I believe this would be the best going forward until #470 is resolved

wolfy1339:
According to https://github.com/github/docs/issues/22726, the payloads will eventually get restored. They have an internal issue tracking this.

katja738 (approved):

katja738 (approved):

ollie-iterators:
https://github.com/github/docs/issues/22726 was closed as not planned 3 weeks ago

wolfy1339:
We unfortunately lose out on some functionality because we are simply downloading the raw HTML, there is only one action listed in the docs since the others are added via JS functionality in the browser.

For new events this shouldn't be too much of a problem as they are added manually.
It would only become a problem if any new actions are added for events because we won't see it in the docs update

gr2m:
@wolfy1339 this PR is way outdated and can be closed, right? Webhooks schemas are now derived from the OpenAPI spec?

wolfy1339:
The update workflow has been broken ever since the GitHub docs switched over to OpenAPI for Webhooks.
I did manage to actually get some updates going.
Probot is close to complete for the new version that switches to OpenAPI, so this repository can be archived when that is done

gr2m:
> Probot is close to complete for the new version that switches to OpenAPI, so this repository can be archived when that is done

Ahh, types for webhooks are not generated directly in https://github.com/octokit/webhooks.js, right?

I think it might still make seens to keep this repository and make it work again in order to support webhook types for the GHES and GHEC. Basically replicate the setup we have for the REST API methods

wolfy1339:
Types for webhooks are generated either in this repository or https://github.com/octokit/openapi-webhooks

Starting with `@octokit/webhooks` v13 it uses the OpenAPI spec.

Probot currently is on v12

wolfy1339:
Since we lost the payloads from the docs, the update script has lost some of it's purpose. It can only return the listed properties

